PR #7, opened 2020-08-05
Initial project layout

mbaynton:


.gitignore
```diff
@@ -1,2 +1,4 @@
 fixtures/tufkeystore
 venv/
+
+/vendor/
```
heddn, line 4:
```suggestion
/vendor/
.idea
composer.lock
.DS_Store
.phpunit.result.cache
```

composer.json
```diff
@@ -0,0 +1,26 @@
...
+    "name": "php-tuf/php-tuf",
+    "description": "PHP implementation of The Update Framework (TUF)",
+    "type": "library",
+    "require-dev": {
+        "phpunit/phpunit": "^4.8",
+        "squizlabs/php_codesniffer": "^3.5"
+    },
+    "license": "MIT",
+    "minimum-stability": "stable",
+    "require": {},
```
heddn, line 11:
```suggestion
    "require": {
        "php": ">=5.3.0"
    },
```

composer.json
```diff
@@ -0,0 +1,26 @@
+{
+    "name": "php-tuf/php-tuf",
+    "description": "PHP implementation of The Update Framework (TUF)",
+    "type": "library",
+    "require-dev": {
+        "phpunit/phpunit": "^4.8",
```
heddn, line 6:
```suggestion
        "phpunit/phpunit": "^4|^5|^6|^7|^8|^9",
        "symfony/phpunit-bridge": "^2|^3|^4|^5"
```

composer.json
```diff
@@ -0,0 +1,26 @@
...
+    "require": {},
+    "autoload": {
+        "psr-4": {
+            "Tuf\\": "src/"
+        }
+    },
+    "scripts": {
+        "code-sniff": [
+            "./vendor/bin/phpcs ./src"
+        ],
```
heddn, line 20:
```suggestion
        "code-sniff": "phpcs ./src ./tests"
```

composer.json
```diff
@@ -0,0 +1,26 @@
...
+        "test": [
+            "./vendor/bin/phpunit tests"
+        ],
```
heddn, line 23:
```suggestion
        "test": "phpunit ./tests"
```

composer.lock
```diff
@@ -0,0 +1,1253 @@
+{
+    "_readme": [
```
heddn, line 2: Remove lock file?

composer.json
xjm, line 6: We discussed that, although we want the client to support down to PHP 5.3 for various projects' legacy versions, we're probably better off writing the tests in a modern, secure, and forward-compatible version of PHPUnit to start. So I'd recommend PHPUnit 8 or 9 here as a dev dependency, and then backport the test suite to PHPUnit 4 as needed later. (We can still support PHP 5 for prod.)

heddn: The issue is that we can't run CI then and certify that we support PHP 5.3+ if we don't also use older versions of PHP Unit

composer.json
```diff
@@ -0,0 +1,25 @@
+{
+    "name": "php-tuf/php-tuf",
+    "description": "PHP implementation of The Update Framework (TUF)",
+    "type": "library",
+    "require-dev": {
+        "phpunit/phpunit": "^4|^5|^6|^7|^8|^9",
+        "symfony/phpunit-bridge": "^2|^3|^4|^5"
```
tedbow: Needs comma

composer.json
```diff
@@ -0,0 +1,25 @@
...
+        "php": ">=5.3.0"
+    },
+    "autoload": {
+        "psr-4": {
+            "Tuf\\": "src/"
+        }
+    },
+    "scripts": {
+        "code-sniff": "phpcs ./src ./tests"
+        "test": "phpunit ./tests"
```
tedbow: needs comma

xjm:
Ted was able install it locally, so merging. Thanks!

PR #8, opened 2020-08-05
CI pipeline

heddn:
We can add PHP 8.0 at a later date. It was acting up on phpunit dependencies (pulling in phpunit 5?!?). For now, let's keep moving and re-look later.


heddn:
We can see that tests are now passing: https://github.com/heddn/php-tuf/actions/runs/196714937

This is ready for review.

PR #12, opened 2020-08-06
Bump minimum PHP to 7.2.

### LINKED ISSUE #6: Minimum system requirements
### PHP Support

- Composer 2 supports back to PHP 5.3
- Typo3 has some other libraries that go back to PHP 5.4

However - should we set our support based on the minimum latest supported version for Typo3, Joomla, or Drupal. 

- Typo3 version 7 - goes back to 5.5 - Typo3 version 8 goes to version 7+ 
- Joomla - minimum version 5.3.10 - recommended PHP 7+
- Drupal 7 - technically supports 5.2, but realistically doesn't support below 5.3 - Drupal 8 and above is in the PHP 7+ 

#### Current Recommendation
 - Let's target a minimum compatibity with PHP 5.3 
 - Note: the cryptographic support is available going this far back using polyfills 
 - Note: We may need test targets for two different versions of PHP to span from 5.3 up to the latest 7/8

heddn:
Fixes #6

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ${{ matrix.operating-system }}
     strategy:
       matrix:
-        operating-system: [ubuntu-latest, macos-latest, windows-latest]
+        operating-system: [ubuntu-latest, macos-latest]
         php-versions: ['7.4', '7.3', '7.2']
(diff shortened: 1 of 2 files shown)


mbaynton:
:green_circle: :100: :1st_place_medal: 

Let's chat with the group about 5.5 in particular. Any particular reason in your mind for that one?

xjm:
Merged #12 to require PHP 7.2.

PR #17, opened 2020-08-06
Add ArrayAccess-based DurableStorage

mbaynton:
This is for abstracting get/set of a TUF client's durable state.

### DIFF
--- a/src/Exception/NotFoundException.php
+++ b/src/Exception/NotFoundException.php
@@ -5,6 +5,9 @@ namespace Tuf\Exception;
 
 use Throwable;
 
+/**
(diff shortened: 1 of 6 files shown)


src/Exception/NotFoundException.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+
+namespace Tuf\Exception;
+
+use Throwable;
+
+class NotFoundException extends \Exception
```
mbaynton, line 8: Looks like I didn't end up using this in the final version of this PR. Happy to remove it, but OTOH we're gonna need things like this pretty soon.

PR #19, opened 2020-08-07
Move to and enforce camelCase vars and params

mbaynton:
Per discussion on the call early this morning. We were being inconsistent. Folks who were around at the time (@tedbow @xjm @mbaynton others?) decided on camelCase because it's probably more familiar to non-Drupal projects and any little niceties we can do for broader adoption are good.

### DIFF
--- a/src/Client/Updater.php
+++ b/src/Client/Updater.php
@@ -31,40 +31,38 @@ class Updater
   /**
    * Updater constructor.
    *
-   * @param string $repository_name
+   * @param string $repositoryName
    * @param array[][] $mirrors
    *   Re
    */
-    public function __construct($repository_name, $mirrors)
+    public function __construct($repositoryName, $mirrors)
     {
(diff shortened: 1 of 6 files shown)


src/Client/Updater.php
```diff
@@ -31,40 +31,38 @@ class Updater
...
         $this->mirrors = $mirrors;
     }
 
 
   /**
    * @todo Update from python comment https://github.com/theupdateframework/tuf/blob/1cf085a360aaad739e1cc62fa19a2ece270bb693/tuf/client/updater.py#L999
-   *
-   * @param bool $unsafely_update_root_if_necessary
    */
-    public function refresh($unsafely_update_root_if_necessary = true)
```
xjm, line 50: Technically out of scope. Can we file a separate issue to refer back to the Python implementation and evaluate whether we need this, and put a `@todo` comment in the code referencing the issue?

PR #23, opened 2020-08-07
Add the Paragon IE polyfill for hosts where ext-libsodium is not installed

xjm:


composer.json
```diff
@@ -12,7 +12,8 @@
     "require": {
         "php": ">=7.2.5",
         "ext-json": "*",
-        "ext-sodium": "*"
+        "ext-sodium": "*",
```
davidstrauss, line 15: Don't we stop requiring `ext-sodium` since we're requiring the polyfill?

xjm: Yep. Fixed in the next commit.

mbaynton:
I think we should be able to officially recommend `ext-sodium` in composer.json as well, which might get eyes on the missing extension in some use cases.

composer.json
```diff
@@ -12,7 +12,7 @@
     "require": {
         "php": ">=7.2.5",
         "ext-json": "*",
-        "ext-sodium": "*"
+        "paragonie/sodium_compat": "^1.13"
     },
```
mbaynton, line 16:
```suggestion
    },
    "suggest": {
        "ext-sodium": "Provides faster verification of updates"
    },
```
I think this should cause composer to print this suggestion for you -- haven't actually tried it w/ a missing extension

xjm: I tested this locally. Note that the paragonie package already outputs the same recommendation. With the addition, it will look like this:
```paragonie/random_compat suggests installing ext-libsodium (Provides a modern crypto API that can be used to generate random bytes.)
paragonie/sodium_compat suggests installing ext-libsodium (PHP < 7.0: Better performance, password hashing (Argon2i), secure memory management (memzero), and better security.)```

Comment (xjm) [reply composer.json:16]: OK, I did more testing. It looks like the correct extension name is `ext-libsodium`, not `ext-sodium`. And apparently Composer is smart enough to not display the message if the package is wrong! Neat. 

With that fixed... Our suggestion will only show up when another package requires ours. I made a test project locally and added `php-tuf/php-tuf` as a dependency:

```
[ayrton:test | Fri 21:50:55] $ composer install
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 3 installs, 0 updates, 0 removals
  - Installing paragonie/random_compat (dev-master 51813fb): Cloning 51813fbc2d from cache
  - Installing paragonie/sodium_compat (v1.13.0): Loading from cache
  - Installing php-tuf/php-tuf (dev-xjm-polyfill): Symlinking from /Users/xjm/git/php-tuf
paragonie/random_compat suggests installing ext-libsodium (Provides a modern crypto API that can be used to generate random bytes.)
paragonie/sodium_compat suggests installing ext-libsodium (PHP < 7.0: Better performance, password hashing (Argon2i), secure memory management (memzero), and better security.)
php-tuf/php-tuf suggests installing ext-libsodium (Provides faster verification of updates)
```

Thoughts?

mbaynton: So three suggestions to install ext-libsodium? Well, my SCRUM trainer liked to tell the class that you have to repeat the same thing three times before adults will remember it. I think it's just because she was covering for not possibly having enough material to fill a few days with otherwise, but it was effective :laughing: 

¯\_(ツ)_/¯ thanks for experimenting with it. I don't think it matters much whether we also include the suggestion or not.

xjm:
OK, let's do this. We can always improve the message later if desired. Thanks!

PR #24, opened 2020-08-08
Require storage implementation in the constructor

mbaynton:
To allow applications to provide their database/filesystem/whatever implementation.

This also updates the tests to use an in-memory copy of the fixtures so
that the test runs are free to write back to the local repo without
creating lasting side-effects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # PHP-TUF
 
-PHP-TUF is a PHP implementation of [The Update Framework(TUF)](https://theupdateframework.io/) to provide signing and verification for secure PHP application updates. [Read the TUF specification](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md) for more information on how TUF is intended to work and the security it provides.
+PHP-TUF is a PHP implementation of [The Update Framework (TUF)](https://theupdateframework.io/) to provide signing and verification for secure PHP application updates. [Read the TUF specification](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md) for more information on how TUF is intended to work and the security it provides.
 
 PHP-TUF project development is primarily focused on supporting secure automated updates for PHP CMSes. Contributing projects:
 - [Drupal](https://www.drupal.org/)
--- a/README.md
+++ b/README.md
@@ -1,31 +1,16 @@
 # PHP-TUF
 
-## What is PHP-TUF?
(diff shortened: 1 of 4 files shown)


src/Client/Updater.php
```diff
@@ -28,21 +28,27 @@ class Updater
    */
     protected $mirrors;
 
-    protected $keyRegistry;
```
mbaynton, line 31: $keyRegistry was never used.

tests/Client/UpdaterTest.php
```diff
@@ -4,9 +4,54 @@
...
 use PHPUnit\Framework\TestCase;
 use Tuf\Client\Updater;
+use Tuf\Tests\DurableStorage\InMemoryBackend;
 
 class UpdaterTest extends TestCase
 {
+    /**
+     * @return Updater
+     */
+    protected function getSystemInTest()
```
xjm, line 14: Should the signature here have a return typehint?

tests/Client/UpdaterTest.php
```diff
@@ -4,9 +4,54 @@
...
+            throw new \RuntimeException("Client repository fixtures directory not found at $path");
+        }
+
+        $storage = new InMemoryBackend();
+
+        $iter = new \FilesystemIterator(
+            $realpath,
+            \FilesystemIterator::SKIP_DOTS | \FilesystemIterator::KEY_AS_FILENAME
+        );
+        foreach ($iter as $filename => $info) {
```
xjm, line 44: `$iter` is not a great variable name. Usually I prefer to use whole words rather than abbreviations. Maybe `$fsIterator` or something?

README.md
```diff
@@ -1,28 +1,31 @@
 # PHP-TUF
```
xjm, line 1: Please ignore changes to this file; they're from a rebase. I thought the PR would automatically redo its diff when rebased against `main`. Sorry!

composer.json
```diff
@@ -12,7 +12,10 @@
     "require": {
         "php": ">=7.2.5",
         "ext-json": "*",
-        "ext-sodium": "*"
```
xjm, line 15: Also ignore these changes; they're in a separate commit from `main`. Sorry!

src/Client/Updater.php
xjm: I think this comment is just an FYI on why the change was made rather than a request for feedback, so marking resolved. Thanks!

PR #26, opened 2020-08-09
Allow CI to pass when a line exceeds 120 characters (but still show the warning locally)

mbaynton:
This is a proposal to be less rigid with style requirements. Not all
observations that phpcs makes are intended by its authors and the
authors of standards like PSR-2/PSR-12 to be hard-and-fast rules that
cannot ever have exceptions. An example is characters per line limits,
which PSR-2 says this about

 > There MUST NOT be a hard limit on line length; the soft limit MUST
   be 120 characters; lines SHOULD be 80 characters or less.

IMO there is value in making warnings available for developers to
consider, (`composer phpcs`), but that this automation does not always
result in better code and contributors should be able to decide not to
comply with every observation phpcs is capable of making.

Any style violation that is a MUST for PSR-2 compliance still fails the build,
and we can elevate the severity of select other warnings if we want.


xjm:
I'm -1 on this. If anything I'd like stricter checking. The advantage of having phpcs enforce standards is that then people don't have to. IMO it greatly improves maintainability to adhere strictly to the standards. This also doesn't seem inconsistent with PSR-2 to me at all. We don't have a hard limit on line length. We should wrap comments, long arrays, etc. at 80 characters for readability.

xjm:
I _would_ however be OK with having two separate jobs with two rulesets. I.e., one that flags warnings for things like overly long lines, and another that fails CI for the main ruleset.

xjm:
OK, I think I found the thing that led to this PR:
https://github.com/php-tuf/php-tuf/pull/34/checks?check_run_id=969330460#step:10:1

I agree that needing to split a one-word string literal and concatenate it when the line is literally only that string literal and necessary indentation is non-ideal. However, rather than throwing the baby out with the bathwater, I think we should look into fine-tuning this a bit by perhaps excluding a specific sub-rule from the primary ruleset rather than suppressing all warnings.

Weirdly, that didn't get flagged for me locally.

tedbow:
I like the idea of explicitly stated the particular rules we want to not fail CI on. I have pushed up change that adds `phpcs-ci-ruleset.xml` that references the existing 1 and just excludes the 1 rule.

xjm:
Another option I thought of in the middle of the night would be to explicitly `// @codingStandardsIgnoreLine` any line that _must_ be over 120 characters because it contains (e.g.) a string literal over 120 chars and so is less readable when split up, but that always means ignoring _all_ the standards for the line. So far phpcs supports ignoring a specific rule for a whole file or all rules for a specific line, so the separate ruleset still seems like the better notion.

xjm:
I pushed https://github.com/php-tuf/php-tuf/pull/40 to test what happens when the line is over 120 chars.

xjm:
#40 passed CI as expected, but the warning doesn't seem to be reported anywhere in GitHub's UI.

Running the phpcs job locally does still report the warning:
```
[ayrton:php-tuf | Wed 08:50:00] $ composer phpcs
> phpcs --standard=phpcs-ruleset.xml ./src ./tests

FILE: /Users/xjm/git/php-tuf/tests/KeyDBTest.php
----------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
----------------------------------------------------------------------
 33 | WARNING | Line exceeds 120 characters; contains 144 characters
----------------------------------------------------------------------

Time: 87ms; Memory: 8MB
```

I verified locally that the `phpcs-ci` job does still fail correctly if a different rule is broken:
```
[ayrton:php-tuf | Wed 08:51:16] $ git diff
diff --git a/src/Client/Updater.php b/src/Client/Updater.php
index 150dc03..4b67c9a 100644
--- a/src/Client/Updater.php
+++ b/src/Client/Updater.php
@@ -57,7 +57,7 @@ class Updater
    */
     public function __construct(string $repositoryName, array $mirrors, \ArrayAccess $durableStorage)
     {
-        $this->repoName = $repositoryName;
+      $this->repoName = $repositoryName;
         $this->mirrors = $mirrors;
         $this->durableStorage = new ValidatingArrayAccessAdapter($durableStorage);
     }
[ayrton:php-tuf | Wed 08:51:21] $ composer phpcs-ci
> phpcs --standard=phpcs-ci-ruleset.xml ./src ./tests

FILE: /Users/xjm/git/php-tuf/src/Client/Updater.php
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 60 | ERROR | [x] Line indented incorrectly; expected at least 8
    |       |     spaces, found 6
----------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------

Time: 79ms; Memory: 8MB

Script phpcs --standard=phpcs-ci-ruleset.xml ./src ./tests handling the phpcs-ci event returned with error code 2
```

xjm:
I wonder if there's any way for GitHub to show the warnings somewhere, without causing CI to fail? @heddn , do you know?

xjm:
Oh wait -- it is possible now to ignore a single rule for a single line!

```
diff --git a/tests/KeyDBTest.php b/tests/KeyDBTest.php
index 47fd7a6..5d8aeb6 100644
--- a/tests/KeyDBTest.php
+++ b/tests/KeyDBTest.php
@@ -30,6 +30,7 @@ class KeyDBTest extends TestCase
                 ],
             1 => [
               "59a4df8af818e9ed7abe0764c0b47b4240952aa0d179b5b78346c470ac30278d",
+              // phpcs:ignore Generic.Files.LineLength.TooLong
               "594e8b4bdafc33fd87e9d03a95be13a6dc93a836086614fd421116d829af68d8f0110ae93e3dde9a246897fd85171455ea53191bb96cf9e589ba047d057dbd66"
             ]
           ]
```

So that is another option. OTOH that adds a lot of overhead every time we have a string that's slightly shorter than a tweet, so I'd still ideally prefer the separate ruleset I think.

xjm:
I'm going to go ahead and merge this to avoid further proliferation of weird string concatenations that get around the line length rule, but still interested in whether it's possible to report the warnings in the UI somehow.

PR #28, opened 2020-08-10
Workflow step 3 WIP

mbaynton:
Highly dependent on #18 to complete all of step 3; this only includes
new things not already implemented there.

Needs tests.


davidstrauss:
This is pretty old. Should we close this?

PR #31, opened 2020-08-10
Improve storage class naming and documentation

xjm:
Improve storage class naming. (Split off from https://github.com/php-tuf/php-tuf/pull/18.)

### DIFF
--- a/src/Client/DurableStorage/FileStorage.php
+++ b/src/Client/DurableStorage/FileStorage.php
@@ -14,10 +14,19 @@ class FileStorage implements \ArrayAccess
 {
     /**
      * @var string $basePath
-     *   The path on the filesystem to this durable storage's files.
+     *     The path on the filesystem to this durable storage's files.
      */
     protected $basePath;
 
+    /**
+     * Constructs a new FileStorage instance.
+     *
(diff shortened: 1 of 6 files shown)


src/Client/DurableStorage/FileStorage.php
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Defines a simple filesystem-based storage for fetched PHP-TUF metadata.
+ *
+ * Applications might want to provide an alternative implementation with
+ * better performance and error handling.
+ *
+ * @todo Add tests for this class.
+ */
+class FileStorage implements \ArrayAccess
```
xjm, line 13: Note that this isn't actually a new file in the PR; it's a file rename that's being detected as a new file due to the extent of changes from the original:
1. Renamed from `FilesystemDurableStorage` to `FileStorage`
2. Protected property `$rootPath` renamed to `$basePath` for disambiguation, with accompanying docs changes
3. protected `offsetToPath()` renamed to `pathWithBasePath()` to better explain what the method actually does
4. Since at that point it's already being detected as a new file rather than a rename, added missing docs and typehints.

PR #32, opened 2020-08-10
Add typed exceptions based on #18

xjm:


mbaynton:
Squash-merging this because the commits that will be lost are not a part of other PRs.

src/Exception/TufException.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Tuf\Exception;
+
+/**
+ * Defines a base class for all exceptions thrown by php-tuf.
+ */
+abstract class TufException extends \Exception
```
mbaynton, line 8: Let's see how this goes with it abstract. I might want to change it back but I'll cite a specific use case if that happens.

PR #34, opened 2020-08-11
Add array formatting rules and fix array indentation

xjm:


tests/KeyDBTest.php
```diff
@@ -21,19 +21,19 @@ public function testComputeKeyId($keyMeta, $want)
...
-            1 => [
-              "59a4df8af818e9ed7abe0764c0b47b4240952aa0d179b5b78346c470ac30278d",
-              "594e8b4bdafc33fd87e9d03a95be13a6dc93a836086614fd421116d829af68d8f0110ae93e3dde9a246897fd85171455ea53191b"
-              . "b96cf9e589ba047d057dbd66"
-            ]
-          ]
+                1 => [
+                    "59a4df8af818e9ed7abe0764c0b47b4240952aa0d179b5b78346c470ac30278d",
+                    "594e8b4bdafc33fd87e9d03a95be13a6dc93a836086614fd421116d829af68d8f0110ae93e3dde9a246897fd85171455e"
+                    . "a53191bb96cf9e589ba047d057dbd66",
```
heddn, line 34: This could be improved.

tests/KeyDBTest.php
```diff
@@ -21,19 +21,19 @@ public function testComputeKeyId($keyMeta, $want)
...
-                  "keyid_hash_algorithms" => ['sha256', 'sha512'],
-                  "keytype" => "ed25519",
-                  "keyval" => ["public" => "edcd0a32a07dce33f7c7873aaffbff36d20ea30787574ead335eefd337e4dacd"],
-                  "scheme" => "ed25519"
+            "case 1" => [
+                0 => [
+                    "keyid_hash_algorithms" => ['sha256', 'sha512'],
+                    "keytype" => "ed25519",
+                    "keyval" => ["public" => "edcd0a32a07dce33f7c7873aaffbff36d20ea30787574ead335eefd337e4dacd"],
+                    "scheme" => "ed25519"
```
heddn, line 29: Should we be testing for trailing comma too?

heddn:
Also, needs rebase

tests/KeyDBTest.php
xjm: The concatenation is necessary for the moment to get CI to pass because of #26, so I think fixing it is out of scope. (It's also concatenated in HEAD, just later because of the missing indent.)

tests/KeyDBTest.php
xjm: Hmm, I thought I added a rule for the trailing comma. Will look into this later.

xjm: So it did add a missing trailing comma in one place where the closing bracket was on a line by itself, but not to this line here. I checked and there is a long open PHPCS issue about this problem: https://github.com/squizlabs/PHP_CodeSniffer/issues/582

PEAR doesn't seem to have anything to say about it either.

So, for this to be detected all the time, we'd need to add some fork of the Drupal array sniff or install another of the various third-party sniffs created related to it.

For now my suggestion is to manually add the trailing comma and file an issue to improve detection for missing trailing commas on multiline arrays.

xjm: I pushed b51bb91 to add the missing trailing comma manually for now.

xjm: Created a followup https://github.com/php-tuf/php-tuf/issues/38 to address this problem, so I propose we resolve this there and fix what we can here, since this can be solved with official sniffs. Thanks!

tests/KeyDBTest.php
xjm: #26 is merged now, so I removed the concatenation.

xjm:
I'm going to go ahead and merge this now because I'm fairly certain both the previous concerns have been addressed (one with a blocker, the other with a followup). Thanks!

PR #37, opened 2020-08-11
Add datetime conversion helper from #18

xjm:
One more small separate scope for review. Soft-blocker for #18.

### DIFF
--- a/src/Client/Updater.php
+++ b/src/Client/Updater.php
@@ -5,6 +5,7 @@ namespace Tuf\Client;
 
 use Tuf\Client\DurableStorage\FilesystemDurableStorage;
 use Tuf\Client\DurableStorage\DurableStorageAccessValidator;
+use Tuf\Exception\FormatException;
 use Tuf\KeyDB;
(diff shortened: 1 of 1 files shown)


src/Client/Updater.php
```diff
@@ -138,6 +139,26 @@ public function validateTarget($targetRepoPath, $targetStream)
...
+     * @param string $timestamp
+     *     The timestamp string in the metadata.
+     *
+     * @return \DateTimeImmutable
+     *     An immutable DateTime object for the given timestamp.
+     *
+     * @throws FormatException
+     *     Thrown if the timestamp string format is not valid.
+     */
+    protected function metadataTimestampToDatetime(string $timestamp) : \DateTimeImmutable
```
tedbow, line 154: I think `Datetime` should be `DateTime`  because this is 2 words and how php functions do this.

xjm: Agreed.

xjm: Pushed a commit to fix this.

tedbow:
Looks good now

PR #39, opened 2020-08-12
Add missing phpdoc and typehints to the JsonNormalizer

xjm:


src/JsonNormalizer.php
```diff
@@ -1,36 +1,45 @@
...
-     *
-     * @return string
      */
-    public static function asNormalizedJson($structure)
+    public static function asNormalizedJson(array $structure) : string
     {
-        if (!is_array($structure)) {
-            throw new \Exception("Array of keys to canonicalize is required");
-        }
-
```
xjm, line 24: Note that the exception is not necessary if we add the typehint to the method signature.

PR #42, opened 2020-08-13
Add a custom phpcs rule to ensure code is documented

xjm:
We're missing a lot of docs right now for parts of the API, which makes PRs like #18 hard to review thoroughly. Squiz has a sniff that _almost_ meets our needs to check for missing docs, but there are two issues with it:
1. It does not support `{@inheritdoc}` (https://github.com/squizlabs/PHP_CodeSniffer/pull/3051)
2. It requires PEAR-like `@param`, `@return`, and `@throws` formatting, which are not very readable and not in the spirit of PSR-2 line length recommendations.

So, this PR adds a custom sniff extending the Squiz sniff for our needs. It's excluded on CI for now until we finish addressing missing docs in the codebase generally.

`@todo` Are we adding the phpcs standard in the best place? It's currently in `TufCs` directory in the project root, but we could move it to a subdirectory if we wanted.


xjm:
> Naming is hard. What we have is probably OK. If desired, another option for the namespace could be `PHP_CodeSniffer\Tuf\Sniffs`. That would follow the pattern of Squiz (in spirit) and get rid of an acronym for code standards (TufCS) in the namespace, which I rarely like.

I've pushed a commit that changes the namespace for the ruleset to `PHP_CodeSniffer\Tuf` and locates the sniffs in ``PHP_CodeSniffer\Tuf\Sniffs`; how's that?

heddn (approved):

PR #44, opened 2020-08-14
Copying docs for classes that were ported from Python TUF

tedbow:


src/KeyDB.php
```diff
@@ -3,6 +3,26 @@
...
+ *
+ * RSA keys are currently supported and a collection of keys is organized as a
+ * dictionary indexed by key ID.  Key IDs are used as identifiers for keys
+ * (e.g., RSA key).  They are the hexadecimal representations of the hash of
+ * key
+ * objects (specifically, the key object containing only the public key).  See
+ * 'rsa_key.py' and the '_get_keyid()' function to learn precisely how keyids
+ * are generated.  One may get the keyid of a key object by simply accessing
+ * the
+ * dictionary's 'keyid' key (i.e., rsakey['keyid']).
```
xjm, line 22: So, we chose to support `ed25519` keys instead of RSA. Is there anything in the code that we'll need to change to account for that, or have we already (so it's just a matter correcting the docs to reflect that)? /cc @mbaynton 

Relatedly, I think it would be a good idea for any place we're doing signing verification to take the encryption strategy as an argument, so other strategies could be supported easily in the future even though the initial implementation doesn't.

tedbow:
No docs were added for RepositoryDBCollection because I think should remove them. see https://github.com/php-tuf/php-tuf/pull/45

but also this wasn't ported from a class in python so we could add the doc later if we want to keep it.

src/KeyDB.php
mbaynton: Yeah we shouldn't talk about RSA support in the comments.

The signature verification function already takes the algorithm type, even though we currently only support one.

mbaynton:
I'm working on a number of edits now. Generally to remove specific references to python classes/modules/variables and references to things we don't yet support.

src/RoleDB.php
```diff
@@ -52,6 +90,31 @@ public function __construct()
         $this->roles = [];
     }
 
+    /**
+     * Add to the role database the 'roleinfo' associated with 'rolename'.
+     *
+     * @param $roleName
+     *     An object representing the role's name, conformant to 'ROLENAME_SCHEMA'
```
mbaynton, line 97: Do you happen to know what `ROLENAME_SCHEMA` is a reference to @tedbow ? References to "conformant to *_SCHEMA" are commonplace in these comments but don't align with anything we currently have. In general I'm wondering how much we should rewrite these or provide definitions for them. I've just referenced sections of the TUF spec in some rewrites in KeyDB, but role names really aren't that complicated and am inclined to just delete that part.

tedbow: Yeah I think we can delete this. I create this issue https://github.com/php-tuf/php-tuf/issues/48
to talk about what schema validation we should have.

src/KeyDB.php
tedbow: Cleaned up the docs including removing these references

src/RoleDB.php
tedbow: Removed the references.

xjm (approved):

PR #47, opened 2020-08-17
Updated fixtures generated with 0.13.0

### LINKED ISSUE #29: Update required version of library in requirements.txt and regenerate test fixtures
`@todo`

mbaynton:
Fixes #29

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,10 @@ code documentation and array formatting.
 
 ### Test fixtures setup
 
-1. [Set up the TUF server environment locally](#server-environment-setup-for-python-tuf-cli).
+1. Install the Python TUF implementation and enable the pipenv:
+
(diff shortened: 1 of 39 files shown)


tedbow:
> Hey @mbaynton. I had two goals here: (1) regenerate using 0.13 and (2) regenerate using target delegation. Since you're using the updated pipenv, the former should already be true. It looks like it doesn't have the latter, though.
> 
> Would you be able to reroll with delegation in place? If you like, we can pair to reroll. It will probably only take 30 minutes.

tedbow:
Sorry, mistake

mbaynton:
If we add some examples of delegation to the fixtures now (we're nowhere near ready to use it), I think we should just be sure to also include some targets that can be validated without delegation.

PR #49, opened 2020-08-19
Make \Tuf\KeyDB::computeKeyIds() private

tedbow:
I am not sure why this method is public. Change to private and changed the test to test `\Tuf\KeyDB::createKeyDBFromRootMetadata()`


davidstrauss (approved):

davidstrauss:
Can we freshen this code from `main` to get the tests passing? I suspect `main` already has the fix.

tedbow:
This was fixed in #95

PR #53, opened 2020-08-22
Unit tests for freeze and rollback attack checks

mbaynton:
@xjm tagged #18 with needs tests. Here are some more.

This moves things around as well though, in ways that I anticipate might provoke discussion (I won't be offended if it _doesn't_...just want to organize topics in case it does.)

The tests here are pure unit tests of the logic that checks for freeze or rollback attacks. Those are methods on the Updater class that are not publicly visible. They should not be made publicly visible to facilitate testing because I care about keeping the public API of the main class our library provides clean. The other popular strategy in PHP circles for unit testing things you don't want public on a given class, putting them in another class and having a whole DI framework to give you access to them, is undesirable I think for this project because we don't have or want a particular whole DI framework in our dependencies.

So, to unit test the methods, I simply extend the Updater class and expose the methods being tested publicly in the subclass. I think this gets the job done, but I don't really see it done in PHP / Drupal.

:warning: Not against main, meant to be merged into #18 to fulfill the needs tests tag.


tests/TestHelpers/DurableStorage/MemoryStorage.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
 
-namespace Tuf\Tests\DurableStorage;
+namespace Tuf\Tests\TestHelpers\DurableStorage;
```
tedbow, line 4: Moving this cause a test failure in DurableStorageAccessValidatorTest. I guess our test only run if the PR is against main. I  will fix

tests/Unit/Client/UpdaterTest.php
```diff
@@ -0,0 +1,94 @@
...
+{
+    use MemoryStorageLoaderTrait;
+
+    protected function getSystemInTest() : Updater
+    {
+        $localRepo = $this->memoryStorageFromFixture('tufclient/tufrepo/metadata/current');
+        return new Updater('repo', [], $localRepo);
+    }
+
+    public function testCheckRollbackAttack_noAttack()
```
tedbow, line 20: The "_" in the method names fails our phpcs. Fixing

tests/PublicVisibility/Client/Updater.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+
+namespace Tuf\Tests\PublicVisibility\Client;
+
+class Updater extends \Tuf\Client\Updater
```
tedbow, line 6: To get rid of this class and future tests classes like this I think it is simpler to do

`$method = new \ReflectionMethod(Updater::class, 'checkRollbackAttack');
  $method->setAccessible(true);
  $method->invoke($sut, $localMetadata, $incomingMetadata);`

This makes it very explicit that testing a non-public method in the test itself.

tedbow (commented):
I am going push the changes for my comments

tedbow (approved):
Looks good. @mbaynton hope you are ok with my change to use `ReflectionMethod` for testing the protected method

PR #57, opened 2020-09-09
Switch fixtures to directly use Python

### LINKED ISSUE #3: Create test fixtures
Create a representative set of TUF test fixtures that can allow a test -- written in PHP against the PHP library -- to verify its correct operation.

We'll want several types of test fixtures:

- [ ] The happy path
- [ ] Unhappy path: expirations
- [ ] Unhappy path: untrusted keys (never trusted)
- [ ] Unhappy path: untrusted keys (replaced by rotation)
- [ ] Unhappy path: manipulated payload

This probably doesn't name every fixture we'll need, but it's a start. We should commit any Python to create the test fixtures to the repository.

davidstrauss:
Fixes #3


bmack (approved):
Makes sense. (see you later hopefully)

PR #65, opened 2020-09-10
Remove RoleDB and use RootMetadata directly

tedbow:


tedbow:
ok. so we get roles for other files. Closing for now

PR #55, opened 2020-09-08
Repo Metadata validation

tedbow:
Against https://github.com/php-tuf/php-tuf/pull/18 for now

PR #66, opened 2020-09-25
Use metadata class files for RoleDb and KeyDb

tedbow:
Currently work off of https://github.com/php-tuf/php-tuf/pull/55


tedbow:
merged into #55 to make it easier to review

PR #67, opened 2020-09-25
Implement updating the root

tedbow:
~~currently off https://github.com/php-tuf/php-tuf/pull/66 which is off  https://github.com/php-tuf/php-tuf/pull/55~~

Against `main` now


src/Client/Updater.php
```diff
@@ -23,6 +23,8 @@
 class Updater
 {
 
+    const MAX_ROOT_DOWNLOADS = 10000000;
```
trishankatdatadog, line 26: Might be a bit excessive 🙂

tedbow: yep. bumped it down to 1024

trishankatdatadog: No problem 🙂 BTW, as an aside, your team and yourself might like to join us to discuss #tuf on the [CNCF Slack](https://cloud-native.slack.com/)! Let me know if you need an invite, just email `trishank.kuppusamy [that-symbol] datadoghq.com`

tedbow: @trishankatdatadog Thanks I just emailed you

src/Client/Updater.php
```diff
@@ -324,4 +307,103 @@ protected function verifySingleSignature(string $bytes, array $signatureMeta)
...
+     *   The role to check for rotated keys.
+     *
+     * @return boolean
+     *   True if the keys for the role have been rotated, otherwise false.
+     */
+    private function hasRotatedKeys(RootMetadata $previousRootData, RootMetadata $newRootData, string $role)
+    {
+        $previousRole = $previousRootData->getRoles()[$role] ?? null;
+        $newRole = $newRootData->getRoles()[$role] ?? null;
+        if (empty($previousRole) && empty($newRole)) {
```
trishankatdatadog, line 404: When would `previousRole` or `newRole` ever be empty? I'm assuming that `previousRole` is based off the currently trusted root metadata on disk, which you should always have, no?

tedbow: Thanks for pointing this out.

I guess I was thinking your could have a root metadata that didn't have a 'timestamp' or 'snapshot' role specified. But I realize now you can't. I see in `4.3. File formats: root.json` of the spec these are required.

I removed this check and added a validation constraint in `\Tuf\Metadata\RootMetadata::getSignedCollectionOptions()` to ensure this is always true.

davidstrauss (approved):
I'd like to comb through the implementation correctness in more detail, but I'd like to do that as part of adding more tests and preparing a release.

Insofar as the overall code architecture and structure (what I really want to check in a PR review), I'm happy and think this is ready to merge.

PR #69, opened 2020-09-28
Fixtures: Add new snapshot/target verification keys, revoke old

### LINKED ISSUE #68: Create test fixtures that rotate the timestamp and snapshot keys to implement and test Spec 1.9
Spec 1.9 starts

> If the timestamp and / or snapshot keys have been rotated, then delete the trusted timestamp and snapshot metadata files.

I don't think our current client fixtures do this because `roles` is the same in `1.root.json` and `3.root.json`

davidstrauss:
Fixes #68

### DIFF
--- a/fixtures/generate.py
+++ b/fixtures/generate.py
@@ -70,3 +70,34 @@ shutil.copytree('tufrepo/metadata.staged/', 'tufrepo/metadata/')
 
 # Generate client metadata
 create_tuf_client_directory("tufrepo/", "tufclient/tufrepo/")
+
+# === Point of No Return ===
+# Past this point, we don't re-export the client. This supports testing the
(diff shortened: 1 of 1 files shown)


tedbow:
I think we should probably just make this PR against #67 because as soon as we generate the the fixtures from the changed script our current test will fail when the updater successfully get the new `4.root.json` from the repo.

davidstrauss:
> I think we should probably just make this PR against #67 because as soon as we generate the the fixtures from the changed script our current test will fail when the updater successfully gets the new 4.root.json from the repo.

Good idea, but how do you feel about just yanking my change into your PR branch?

tedbow:
@davidstrauss sounds good will add this my PR

PR #71, opened 2020-10-06
Add RemoteRepoFileFetcherInterface to fetch remote repo files

tedbow:


src/Client/RemoteRepoFileFetcherInterface.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace Tuf\Client;
+
+/**
+ * Defines an interface for fetching remote repo files.
+ */
+interface RemoteRepoFileFetcherInterface
```
davidstrauss, line 8: Why not just `RepoFileFetcherInterface`? Couldn't we implement the fetching against local data sources equivalently, even just for testing?

davidstrauss (approved):
Other than my extremely minor bike shedding (naming) comment, this looks great.

src/Client/RemoteRepoFileFetcherInterface.php
tedbow: Sounds good. Changed the name.

Issue #70, opened 2020-10-02
Add back phpcs testing to PHP 8 when phpcs version 3.5.7 is released


tedbow:
Our composer `phpcs-ci` script is failing only for php8 https://github.com/php-tuf/php-tuf/runs/1099036890

2 errors

>   75 | ERROR | Type hint "\ArrayAccess" missing for $durableStorage
>   229 | ERROR | Type hint "\DateTimeInterface" missing for $now

But these are both present

tedbow:
from #73 Looks like the latest dev version of phpcs will fix this. 

There are a lot of commits on phpcs since 3.5.6 that mention "PHP 8.0". Assuming one fixes this. We could do #72 now to remove the errors until phpcs 3.5.7 comes out or we could just put up with the errors until then. 

cc @ericduran since you mentioned you were going to look at this.

PR #72, opened 2020-10-07
remove phpcs from running for php8

tedbow:
I created https://github.com/php-tuf/php-tuf/issues/70 to figure out what is going on with phpcs and php 8. The failure looks like a bug.

I think it is better to stop this check for now so there is no confusion about checks on currently supported version of php passing. The actual phpunit tests will still run on PHP 8 and phpcs will run on all versions for PHP 7 that are in our matrix.

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -58,6 +58,7 @@ jobs:
         run: composer lint
 
       - name: Run PHPCS
+        if: matrix.php != '8.0'
         run: composer phpcs-ci
 
       - name: Run test suite


tedbow:
#73 proves the next phpcs release should fix this. I don't think we should use the dev version for now.

So we can do this PR now or just accept the failures until phpcs 3.5.7 comes out.

davidstrauss:
Is there anything like an XFAIL we can use here?

tedbow:
> Is there anything like an XFAIL we can use here?

@davidstrauss hmmm, that would better but I couldn't find anything in github workflows that would allow use to that.

I created #70 as todo to add remove this condition when the new version is released.

davidstrauss (approved):
If we can't XFAIL, then I agree that what you've done is the next best thing.

PR #73, opened 2020-10-07
use dev-master for phpcs see if this fixes PHP 8 bug

tedbow:


tedbow:
This was just a test to see phpcs dev-master would solve our problem which it does

PR #75, opened 2020-10-30
Implementing the TUF spec section 5.3

tedbow:
Implementing the TUF spec section 5.3
https://github.com/theupdateframework/specification/blob/v1.0.9/tuf-spec.md#5-detailed-workflows

Instead of 1 set of fixtures we now have `fixture/delegated` and `fixtures/simple`. `fixture/delegated` is the same as the existing fixtures. I added the new `fixtures/simple` because `fixture/delegated` does not allow testing comparing the new timestamp/snapshot files to the current trusted ones. When the root is update and the timestamp/snapshot keys are rotated the trusted version are discarded.
Obviously looking at the changes to `fixtures/generate.py`(be kind my first python code changes ever 😉 ) is easier than inspecting all of the fixture files.

I changed some other things along the way such as switching to static methods when possible.


davidstrauss (approved):

davidstrauss:
This looks great, and I agree with the decision.

PR #77, opened 2020-11-10
Add stub for meeting minutes to the repository

xjm:


davidstrauss:
Could we track these outside of the main git repo, possibly? What do you think of using the GitHub wiki functionality, which also stores in git (but not the code git repo)?

davidstrauss (changes requested):
Could we track these outside of the main git repo, possibly? What do you think of using the GitHub wiki functionality, which also stores in git (but not the code git repo)?

hestenet (approved):

hestenet:
> Could we track these outside of the main git repo, possibly? What do you think of using the GitHub wiki functionality, which also stores in git (but not the code git repo)?

Actually - this is probably a better idea! I'm going to play around with that.

hestenet:
Yeah, the wiki will work well, so I'm going to close this PR

Work will continue here: https://github.com/php-tuf/php-tuf/wiki

xjm:
Works for me. I added a link to the wiki to the README as well. (I know it's in a tab, but still better for the README to be the central starting point for all docs IMO).

PR #79, opened 2020-11-10
Generate test fixtures in GitHub Actions

davidstrauss:


davidstrauss:
We'll definitely want to squash this when merging, as the only way I could test certain things on GitHub actions was by trial and error pushing commits.

tedbow (approved):
This looks good. I downloaded the fixtures artifact from the test run compared to `main` only the keys, sig and expires changed as expected.

PR #80, opened 2020-11-10
Implements 5.4.1 to 5.4.4: Download and verify the top-level targets metadata file

tedbow:
Implements 5.4 to 5.4.4 which is similar to the logic for the snashot.json file in 5.3.x. 

The logic in 5.4.5 is targets specific so my plan was to do 5.4.5 to 5.4.5.2.3(the last step in 5.4) in the next PR


src/Client/Updater.php
```diff
@@ -163,14 +165,14 @@ public function refresh() : bool
                 "$snapShotVersion.snapshot.json",
                 static::MAXIMUM_DOWNLOAD_BYTES
             );
+            // TUF-SPEC-v1.0.9 Section 5.3.1
```
xjm, line 168: Given the fact that the numbering does change, I wonder if we should:
- Simply link the relevant spec item
- Say in words what the step is (the bolded part in the spec doc)

tests/Client/TestRepo.php
```diff
@@ -62,15 +64,19 @@ public function fetchFile(string $fileName, int $maxBytes)
     }
 
     /**
-     * Sets the file for which a signature fail should be triggered.
+     * Sets a value in a file to be changed in ::fetchFile().
```
xjm, line 67: I don't understand the new purpose of this method from these docs.

tedbow: The class is a very convoluted now. I will change so that this method changes the repo files directly instead of storing the change information for later.

src/Client/Updater.php
```diff
@@ -398,31 +420,36 @@ private function updateRoot(RootMetadata &$rootData)
         $originalRootData = $rootData;
         // *TUF-SPEC-v1.0.9 Section 5.1.2
         $nextVersion = $rootData->getVersion() + 1;
-        while ($nextRootContents = $this->repoFileFetcher->fetchFile("$nextVersion.root.json", static::MAXIMUM_DOWNLOAD_BYTES)) {
-            $rootsDownloaded++;
-            if ($rootsDownloaded > static::MAX_ROOT_DOWNLOADS) {
-                throw new \Exception("The maximum number root files have already been dowloaded:" . static::MAX_ROOT_DOWNLOADS);
+        try {
+            while ($nextRootContents = $this->repoFileFetcher->fetchFile("$nextVersion.root.json", static::MAXIMUM_DOWNLOAD_BYTES)) {
```
tedbow, line 424: This nesting is now need because TestRepo was not implementing `\Tuf\Client\RepoFileFetcherInterface::fetchFile` correctly.

tests/Client/TestRepo.php
```diff
@@ -34,43 +29,41 @@ class TestRepo implements RepoFileFetcherInterface
...
-            // exception.
-            // @see \Tuf\Client\Updater::checkSignatures()
-            if (in_array($fileName, $this->failSignatureFiles)) {
-                $json = json_decode($contents, true);
-                $json['signed']['extra_test_value'] = 'value';
-                $contents = JsonNormalizer::asNormalizedJson($json);
-            }
-            return $contents;
-        } catch (\Exception $exception) {
-            return false;
```
tedbow, line 60: It was suppose to throw an exception if the file is not found but it was returning false. Obviously we are at the point where we could change this if we wanted to but I think throwing the exception is good.

tests/Client/UpdaterTest.php
```diff
@@ -63,27 +93,25 @@ protected function getSystemInTest() : Updater
...
-     *   The expected start versions.
      * @param array $expectedUpdatedVersions
      *   The expected updated versions.
      *
      * @return void
      *
      * @dataProvider providerRefreshRepository
      */
-    public function testRefreshRepository(string $fixturesSet, array $expectedStartVersions, array $expectedUpdatedVersions) : void
+    public function testRefreshRepository(string $fixturesSet, array $expectedUpdatedVersions) : void
```
tedbow, line 103: I removed `$expectedStart` version because this doesn't vary with every test case but instead with the feature set used. So this can be determined from `$fixturesSet`.

Added `getFixtureClientStartVersions()` so now every test in the class can assert the starting point based on the feature set.

tests/Client/UpdaterTest.php
```diff
@@ -97,123 +125,257 @@ public function providerRefreshRepository()
...
-     * Asserts that files in the repo are at expected versions.
+     * Asserts that files in the client repo are at expected versions.
      *
      * @param array $expectedVersions
      *   The expected versions.
      *
      * @return void
      */
-    protected function assertRepoVersions(array $expectedVersions): void
+    protected function assertClientRepoVersions(array $expectedVersions): void
```
tedbow, line 155: Updated the name and comment here to make it clear we are asserting the client repo state

tests/Metadata/MetaDataBaseTest.php
```diff
@@ -285,28 +285,6 @@ public function testInvalidField(string $expectedField, string $expectedType) :
...
-    protected function nestedChange(array $keys, array &$data, $newValue) : void
-    {
-        $key = array_shift($keys);
-        if ($keys) {
-            $this->nestedChange($keys, $data[$key], $newValue);
-        } else {
-            $data[$key] = $newValue;
-        }
-    }
-
```
tedbow, line 309: No change this method just moved to existing `UtilsTrait` so it can be used in `TestRepo`

tests/Client/UpdaterTest.php
```diff
@@ -97,123 +125,257 @@ public function providerRefreshRepository()
...
+     * @param array $expectedUpdatedVersions
+     *   The expected repo file version after refresh attempt.
      *
      * @return void
      *
-     * @dataProvider providerSignatureError
+     * @dataProvider providerRefreshException
      */
-    public function testSignatureError(string $fileToFail, int $expectedRootVersion, string $expectionMessage) : void
+    public function testRefreshException(string $fileToChange, array $keys, $newValue, \Exception $expectionException, array $expectedUpdatedVersions): void
```
tedbow, line 189: Instead of a specific test for signature errors I changed this into generic method to test different types of exceptions when files are changed.

src/Client/Updater.php
```diff
@@ -398,31 +420,36 @@ private function updateRoot(RootMetadata &$rootData)
...
-            // Note that the expiration of the new (intermediate) root metadata
-            // file does not matter yet, because we will check for it in step
-            // 1.8.
-
-            // *TUF-SPEC-v1.0.9 Section 5.1.6 and 5.1.7
-            $this->durableStorage['root.json'] = $nextRootContents;
-            $nextVersion = $rootData->getVersion() + 1;
-            // *TUF-SPEC-v1.0.9 Section 5.1.8 Repeat the above steps.
+        } catch (RepoFileNotFound $exception) {
+            // This exception is expected when the newest root file has already been downloaded.
```
phenaproxima, line 450: I'm not sure I would do this; it feels like a misuse of an exception, and it might cover up some sort of freak error condition (maybe the file DOES exist, but the server is returning a 404 due to some error condition). IMHO we should just add a `fileExists()` method to `RepoFileFetcherInterface` to handle this case, so we don't try to download a non-existent file as a way of verifying its existence in the first place.

trishankatdatadog: Sorry, not sure I understand. How else would you check whether the file exist or not on the repo?

phenaproxima: In the case of the GuzzleFileFetcher implemented by #81, I was thinking maybe a `HEAD` request would do the trick, instead of `GET` (which will be heavier because it'll return the full content of the file). Using a lighter-weight request in order to check existence also gives us some wiggle room to automatically retry a few times before giving up.

On the other hand, I could be overthinking it too :)

trishankatdatadog: I think a HEAD request, and then, if successful, followed by a GET request makes sense 🙂

tedbow: @phenaproxima hadn't seen your last comment but was thinking of adding 
`/**
     * Gets a file if it exists if it exists in the remote repo.
     *
     * @param string $fileName
     *   The file name to fetch.
     *
     * @return string|null
     *   The contents of the file or null if it does not exist.
     */
    private function fetchFileIfExists(string $fileName):?string
    {
        try {
            return $this->repoFileFetcher->fetchFile($fileName, static::MAXIMUM_DOWNLOAD_BYTES);
        } catch (RepoFileNotFound $exception) {
            return null;
        }
    }`

Then we would need the try/catch around the loop. This was the `Updater` class but could be for `RepoFileFetcherInterface`

tedbow: add to RepoFileFetcherInterface

tests/Client/TestRepo.php
tedbow: Changed

src/Client/Updater.php
xjm: Let's ignore this in this PR and fix later once we have a response to https://github.com/theupdateframework/specification/pull/133.

trishankatdatadog:
Good job, guys, I'm here myself with [tuf-on-a-plane](https://github.com/trishankatdatadog/tuf-on-a-plane/pull/17)

tedbow:
@trishankatdatadog I think I remember what a plane is 😂

src/Client/Updater.php
```diff
@@ -186,6 +189,25 @@ public function refresh() : bool
...
+                "$targetsVersion.targets.json",
+                static::MAXIMUM_DOWNLOAD_BYTES
+            );
+            // TUF-SPEC-v1.0.9 Section 5.4.1
+            $this->confirmFileFromExistingMetadata($newSnapshotData, $newTargetsContent);
+            $newTargetsData = TargetsMetadata::createFromJson($newTargetsContent);
+            // TUF-SPEC-v1.0.9 Section 5.4.2
+            $this->checkSignatures($newTargetsData);
+            // TUF-SPEC-v1.0.9 Section 5.4.2
+            static::checkFreezeAttack($newTargetsData, $nowDate);
```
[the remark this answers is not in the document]
tedbow: I used " 5.4.2" 2x. fixed the 2nd comment

phenaproxima (approved):

src/Client/Updater.php
```diff
@@ -185,6 +185,25 @@ public function refresh() : bool
         // TUF-SPEC-v1.0.9 Section 5.3.5
         $this->durableStorage['snapshot.json'] = $newSnapshotContents;
 
+        // TUF-SPEC-v1.0.9 Section 5.4
+        if ($rootData->supportsConsistentSnapshots()) {
+            $targetsVersion = $newSnapshotData->getFileMetaInfo('targets.json')['version'];
+            $newTargetsContent = $this->repoFileFetcher->fetchFile(
+                "$targetsVersion.targets.json",
+                static::MAXIMUM_DOWNLOAD_BYTES
```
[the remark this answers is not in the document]
tedbow: Good catch! Right ow our fixtures don't have the`length` element in `snapshots.json`. We figure out if we can test fixtures with this or decide on support. 

Created a meta issue to figure out handling of variations not covered in current test fixtures and code https://github.com/php-tuf/php-tuf/issues/97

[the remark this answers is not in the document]
tedbow: Just merged #99 which will throw an exception if `hash` or `length` or are specified here. The are optional fields that aren't supported by the Python library.

Importantly though the `targets.json` file itself that we download here is signed by trusted keys so even though we don't have the `hashes` specified in `snapshot.json` we still are verifying the contents of the metadata in the file.

src/Client/Updater.php
```diff
@@ -185,6 +185,25 @@ public function refresh() : bool
...
+            );
+            $newTargetsData = TargetsMetadata::createFromJson($newTargetsContent);
+            // TUF-SPEC-v1.0.9 Section 5.4.1
+            $newSnapshotData->verifyNewMetaData($newTargetsData);
```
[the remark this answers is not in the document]
davidstrauss: +1 on this thought. Should we also consider a tainting model (where the taint gets removed on verification and accessing tainted data requires acknowledgement that it's not known to be safe)? Maybe that's overkill, though.

tests/Client/UpdaterTest.php
```diff
@@ -302,6 +303,18 @@ public function providerRefreshException()
...
+                '5.targets.json',
+                ['signed', 'version'],
+                6,
+                new MetadataException("Expected targets version 5 does not match actual version 6."),
+                [
+                    'root' => 5,
+                    'timestamp' => 5,
+                    'snapshot' => 5,
+                    'targets' => 3,
+                ],
```
davidstrauss, line 316: Is there a way to avoid duplicating test condition data in the exception? Could we carry over expected/actual data in a more modular fashion?

davidstrauss (changes requested):

src/Client/Updater.php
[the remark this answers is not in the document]
tedbow: @davidstrauss ok I added `isTrusted()` for certain getters and added an optional parameter `$allowUntrustedAccess` when sometimes you would need to access this information before the metadata was verified. 

For getters like `getSigned()` and `getVersion()` I didn't add this because we always need to access this before it is signed.

cc @phenaproxima

tests/Client/UpdaterTest.php
tedbow: I don't follow what you mean here. Have the exception here in the test case allows using the same `testRefreshException()` test method for any cases where an exception should occur based on change to specific repo file.

I am up for changing this to be cleaner but since it is existing data provider I would rather do it in separate PR. Could be before or after this 1.

davidstrauss: Okay. I'll switch my review to "approved" based on this. I think I misunderstood how this was integrating with the rest of the code.

src/Client/Updater.php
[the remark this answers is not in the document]
davidstrauss: Very cool!

davidstrauss (approved):

PR #81, opened 2020-11-11
Add Guzzle as a dependency and implement a file fetcher that uses it

phenaproxima:


tests/Unit/GuzzleFileFetcherTest.php
```diff
@@ -0,0 +1,90 @@
...
+        $history = [];
+        $handlerStack->push(Middleware::history($history));
+        $client = new Client(['handler' => $handlerStack]);
+
+        $fetcher = new GuzzleFileFetcher($client);
+        if ($exceptionClass) {
+            $this->expectException($exceptionClass);
+        }
+        $data = $fetcher->fetchFile('test.json', $maxBytes);
+        $this->assertSame($content, $data);
```
tedbow, line 72: This seems like it could be just 2 test methods. For all the cases where there is exception the last assert is never called. So maybe one for requests with exceptions and 1 without.

everything above this segment code be refactor into a common function. Like `\Tuf\Tests\Unit\Client\UpdaterTest::getSystemInTest()` because are using that pattern multiple places in our tests

src/Client/GuzzleFileFetcher.php
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Creates an instance of this class with a specific base URI.
+     *
+     * @param string $baseUri
+     *   The base URI from which to fetch files.
+     *
+     * @return static
+     *   A new instance of this class.
+     */
+    public static function createFromUri(string $baseUri) : self
```
tedbow, line 44: Is there a reason we need this to be public except for the tests? Won't we expect most cases to use `createFromUri()`?  Or do we want to support the use of passing in something other than `\GuzzleHttp\Client`?

src/Client/GuzzleFileFetcher.php
```diff
@@ -0,0 +1,84 @@
...
+    {
+        $scheme = parse_url($baseUri, PHP_URL_SCHEME);
+        if ($scheme === 'https') {
+            $client = new Client(['base_uri' => $baseUri]);
+            return new static($client);
+        } else {
+            throw new \InvalidArgumentException("Repo base URI must be HTTPS: $baseUri");
```
tedbow, line 51: Not sure if we should allow non `https`. I know @davidstrauss has mention that using TUF we don't have to trust the HTTPS.

src/Client/GuzzleFileFetcher.php
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fetchFile(string $fileName, int $maxBytes)
+    {
+        try {
+            $response = $this->client->request('GET', $fileName);
+        } catch (RequestException $e) {
```
tedbow, line 62: Could we change this to `\GuzzleHttp\Exception\ClientException` it says it covers "4xx codes"

tests/Unit/GuzzleFileFetcherTest.php
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * Tests creating a file fetcher with a repo base URI.
+     *
+     * @return void
+     *
+     * @covers ::createFromUri
+     */
+    public function testCreateFromUri(): void
```
tedbow, line 82: Should we have a test for a call with no exception?

tedbow (commented):
Look good. Mostly just some questions

tests/Unit/GuzzleFileFetcherTest.php
```diff
@@ -0,0 +1,90 @@
...
+     * Data provider for testFetchFile().
+     *
+     * @return array[]
+     *   Sets of arguments to pass to the test method.
+     */
+    public function dataProvider(): array
+    {
+        return [
+            [404, 256, RepoFileNotFound::class],
+            [403, 256, 'RuntimeException'],
```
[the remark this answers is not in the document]
phenaproxima: > why not use ::class in both case?

Because the fully qualified name of \RuntimeException is 'RuntimeException', and adding `::class` would be redundant, extra typing. I don't feel strongly about it, though, so I'll fix it if you want.

> Also we could just use new RuntimeException('the message') and then assert the class and messages match. Especially because we are using RuntimeException be good to check it is actually the one we are expecting.

The problem with this is that I don't know what error message will be generated by Guzzle for 403s, 500s, or other non-404 codes. And the file fetcher doesn't really need to care about that. How about, if we are expecting a \RuntimeException, I call `$this->expectExceptionCode($statusCode)` instead? That would confirm that the correct exception is being thrown.

src/Client/GuzzleFileFetcher.php
phenaproxima: `createFromUri()` is the actual entry point I expect calling code to use. The only reason the constructor is public is to facilitate unit testing.

src/Client/GuzzleFileFetcher.php
phenaproxima: We don't have to, but maybe we should? It seems like a reasonable restriction, at least for now.

src/Client/GuzzleFileFetcher.php
phenaproxima: How about \GuzzleHttp\Exception\BadResponseException instead? That'll cover 5xx codes too.

tests/Unit/GuzzleFileFetcherTest.php
phenaproxima: We do. First line of `testCreateFromUri()`:
```
GuzzleFileFetcher::createFromUri('https://example.com');
```

tests/Unit/GuzzleFileFetcherTest.php
phenaproxima: > This seems like it could be just 2 test methods. For all the cases where there is exception the last assert is never called.

For the record, that was totally intentional on my part :) But I agree with you that it's kinda implict non-obvious programming, so I'm okay with splitting this into two methods.

tests/Unit/GuzzleFileFetcherTest.php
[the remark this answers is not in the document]
tedbow: could we add just always call `$this->expectExceptionCode($statusCode)`.  everything else seems fine the way it is

tests/Unit/GuzzleFileFetcherTest.php
tedbow: obvious seems better. So lets split it

src/Client/GuzzleFileFetcher.php
tedbow: I think it would nice if we could make the constructor private if we don't see it as an entry point. here is an idea on that https://github.com/tedbow/php-tuf/pull/1/commits/89df373b2da4b68cc68b8bdf26848aa5e5264c7d#diff-22262933f442133c3659f5b829b1cb3cd89362e42f52b48474c514eacfa99aaaR71

src/Client/GuzzleFileFetcher.php
tedbow: ok fine leaving as is until we get a request for this(and maybe rejecting the request)

src/Client/GuzzleFileFetcher.php
tedbow: but we don't care about 5xx codes. We are just going to rethrow anyways. So seems like we should be as narrow as we can. If they had a 404Except I would say you use that but there isn't one

tests/Unit/GuzzleFileFetcherTest.php
tedbow: sorry read to fast. We don't assert the class of the return

src/Client/GuzzleFileFetcher.php
phenaproxima: I'd prefer not to do that, if I can, because it's "dishonest" -- it also couples the test to the _internal_ implementation of the class, rather than strictly testing what it _does_. The pattern I already have is basically plain dependency injection. So...are you sure? :)

tedbow: It's fine

tests/Unit/GuzzleFileFetcherTest.php
tedbow: I meant to say lets assert the return value. like
`$this->assertInstanceOf(GuzzleFileFetcher::class, GuzzleFileFetcher::createFromUri('https://example.com'));`
otherwise we just know it didn't throw an exception. it still could return null or false.

tedbow (approved):
All my changes were addressed. Looks good

PR #83, opened 2020-11-11
Deterministic fixtures

davidstrauss:
By having each test fixture repository generated using a fixed "current time" and "generating" key pairs from a predefined list, always in the same order, we can get test fixtures that generate flexibly and deterministically.


tedbow (approved):
Looks good to me! Like the refactoring into python classes

Issue #61, opened 2020-09-10
Use typed exceptions


xjm:
`src/Exception` contains a `TufException` base class and typed exceptions for various circumstances. However, there are several places where `\Exception` is used, including:
- `KeyDB`
- `RoleDB`
- A couple of places in `Updater`

Replace these with appropriate typed exceptions.

tedbow:
I am having trouble assigning right now but @ajfwebdev is interested in working on this.

@ajfwebdev we have doing this little by little. Not sure it still needs anything. 

But if you would look through the exceptions thrown and see if any of the ones not in the php-tuf namespace need to be replaced.

I see 1 `throw new \Exception('Role already exists: ' . $role->getName());` that should replaced with a typed exception. but for one like  
`throw new \RuntimeException('\Tuf\JsonNormalizer::rKeySort() is not intended to sort objects except \ArrayObject. Found: ' . get_class($structure));` like that is not related to an attack probably don't need to be replaced.

Issue #21, opened 2020-08-07
Determine if "update root unsafely necessary" flag is needed for updater


xjm:
The [Python client's updater refresh method](https://github.com/theupdateframework/tuf/blob/1cf085a360aaad739e1cc62fa19a2ece270bb693/tuf/client/updater.py#L999) has an optional `unsafely_update_root_if_necessary` boolean argument that defaults to TRUE.

We're currently not using it. Is it needed?

joshuagl:
`unsafely_update_root_metadata` is part of the trust bootstrapping process. The specification assumes 
>that a good, trusted copy of this file was shipped with the package manager or software updater using an out-of-band process.

the first part of the [detailed client workflow](https://github.com/theupdateframework/specification/blob/c6ec4fc0d8f5f32bff464aca75b153f373d6bc82/tuf-spec.md#5-detailed-workflows) is to start with the trusted metadata and update to the latest version from the repository, verifying that each new version (n+1) of the root metadata is signed by a threshold of signatures from the trusted metadata (n) and itself (n+1).

Because this out of band shipped root metadata may be expired, we need a way to load a file that would usually fail verification. Hence the `unsafely_update_root_metadata` argument.

mbaynton:
Makes sense to me.

PR #84, opened 2020-11-12
Use more specific exceptions where appropriate

phenaproxima:
Addresses #61.


mbaynton:
My thoughts on this one were requested. Sorry my attention was pre-empted while discussing on the call, but for me it boils down to

- Throwing `TufException`s as much as possible / essentially always would be a nice goal for the advantage that users of the library would be able to easily isolate the cause of failed updates to php-tuf.
- That said, I don't personally mind breaking with that goal and throwing normal predefined exception types such as `\RuntimeException` for occurrences that arise from programming errors and not from unexpected or otherwise problematic inputs. That looks like all these particular changes do, so I'm +1 on them.
- Other exception types, in particular from Guzzle, were brought up as a reason we wouldn't be able to constrain our exceptions due to unexpected or otherwise problematic inputs to `TufException`s. If we've gone ahead with adding Guzzle as a dependency, my 2 cents is it'd be cool if we could cover common exceptions Guzzle might throw and rethrow those as TufExceptions, as this might help consumers of our library to integrate it successfully.

src/KeyDB.php
```diff
@@ -151,15 +152,15 @@ public function addKey(array $keyMeta)
...
+     * @throws \InvalidArgumentException
      *     Thrown if the key ID is not found in the keydb database.
      *
      * @see https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#4-document-formats
      */
     public function getKey(string $keyId)
     {
         if (empty($this->keys[$keyId])) {
-            throw new \Exception("Unknown key ID: $keyId");
+            throw new \InvalidArgumentException("Unknown key ID: $keyId");
```
mbaynton, line 163: Hmm, actually I'm not sure about this one. Key IDs are often read from other TUF repository metadata, so I sort of expect that a malformed remote TUF repository that referenced a nonexistent key would end up throwing here. To my mind that's in the class of bad input problems, not programming problems.

mbaynton: Likewise with the unknown role name. What about a `TufException` subclass for all malformed TUF repository issues?

phenaproxima: I ended up using the pre-existing `\Tuf\Exception\NotFoundException`, which seems to be the right fit for these errors.

mbaynton (approved):

src/JsonNormalizer.php
```diff
@@ -33,15 +33,15 @@ public static function asNormalizedJson(array $structure) : string
...
+     * @throws \RuntimeException
      *     Thrown if sorting the array fails.
      *
      * @return void
      */
     private static function rKeySort(array &$structure) : void
     {
         if (!ksort($structure, SORT_STRING)) {
-            throw new \Exception("Failure sorting keys. Canonicalization is not possible.");
+            throw new \RuntimeException("Failure sorting keys. Canonicalization is not possible.");
```
xjm, line 44: So this lightweight JSON canonicalization that we're implementing is a source of problems (exactly related to what David and Ted were debugging today), and this seems like a way to insert a vulnerability. "Feed it garbage JSON that causes a sort error, and then halt their updater." I feel we might want to step up a level at this whole API is a separate issue.

src/RoleDB.php
```diff
@@ -61,67 +62,12 @@ public static function createFromRootMetadata(RootMetadata $rootMetadata)
...
-     *     The role name.
-     *
-     * @return boolean
-     *     True if the role is found in the role database; false otherwise.
-     */
-    public function roleExists(string $roleName)
-    {
-        return !empty($this->roles[$roleName]);
-    }
-
```
xjm, line 124: I'd rather remove this any other such dead code in a separate, dedicated issue, with code commends and followup issues to revisit it (similar to https://github.com/php-tuf/php-tuf/issues/21 ).

src/RoleDB.php
```diff
@@ -163,25 +109,4 @@ public function getRoleKeyIds(string $roleName)
...
-     *     Thrown if the role does not exist.
-     */
-    public function getRoleThreshold(string $roleName)
-    {
-        if (! $this->roleExists($roleName)) {
-            throw new \Exception("Role does not exist: $roleName");
-        }
-
-        return $this->roles[$roleName]['threshold'];
-    }
```
xjm, line 186: More mostly dead code (to discuss separately as above).

xjm (commented):
The rest looks pretty good; I like where this is going.

tedbow (changes requested):
This needs to merged with main. Also I think the remove of code should be in separate issue. I think the removal would cause failures

src/RoleDB.php
phenaproxima: Restored all this.

src/RoleDB.php
phenaproxima: Restored for now.

phenaproxima:
OK, I merged `main` back into this branch and restored (for now) the dead code, for removal in a different issue.

xjm:
Can we file said different issue for dead code removal before merging this?

phenaproxima:
Filed #102 for that.

src/RoleDB.php
```diff
@@ -16,7 +17,7 @@ class RoleDB
      *
      * @var \array[]
      */
-    protected $roles;
+    protected $roles = [];
```
tedbow, line 20: Unrelated change

src/RoleDB.php
```diff
@@ -61,7 +62,7 @@ public static function createFromRootMetadata(RootMetadata $rootMetadata)
             }
             $roleInfo['paths'] = [];
 
-            $db->addRole($roleName, $roleInfo);
+            $db->roles[$roleName] = $roleInfo;
```
tedbow, line 65: Unrelated change

tedbow (changes requested):

tedbow (approved):
Looks good

PR #85, opened 2020-11-12
Used datetime validator instead of regex for 'expires' validation

tedbow:


tests/Metadata/MetaDataBaseTest.php
```diff
@@ -126,7 +126,7 @@ public function testExpires(string $expires, bool $valid) : void
         $metadata['signed']['expires'] = $expires;
         if (!$valid) {
             $expectedMessage = preg_quote('Array[signed][expires]', '/');
-            $expectedMessage .= '.*This value is not valid.';
+            $expectedMessage .= '.*This value is not a valid datetime.';
```
tedbow, line 129: This a plus that we get a more specific message

tests/Metadata/MetaDataBaseTest.php
```diff
@@ -315,19 +315,19 @@ protected function nestedChange(array $keys, array &$data, $newValue) : void
      */
     public function providerExpires() : array
     {
-        return [
+        return static::getKeyedArray([
             ['1970', false],
-            ['1970-01-01T00:00:00Z', false],
+            ['1970-01-01T00:00:00Z', true],
```
tedbow, line 320: A couple test cases now pass because we don't have a hard-code "2" to start the year. Think that is fine.

Just means our descents won't have have mad rush to fix this is 2999 😉

davidstrauss (approved):

phenaproxima (approved):

Issue #60, opened 2020-09-09
Add getter methods for different properties under "signed" for different *MetaData classes.


tedbow:
Currently in https://github.com/php-tuf/php-tuf/pull/55
There are getters on MetadataBase to get the properties like 'version' and 'expires' that are common across all metadata types. 

Each metadata type will need specific getters for properties that only are applicable for that type. Like 'roles' for RootMeta data. This will make the classes more useful. For instance we could pass a RootMeta class to RolesDB.

This will probably a PR for each *MetaData class

phenaproxima:
Since we now require PHP 8.1, I'd suggest we use `public readonly` properties were possible. Much more convenient!

phenaproxima:
We can close this out as of #367; everything in `signed` can be accessed as a public property, but not modified thanks to `readonly`.

PR #86, opened 2020-11-12
Add specialized getters for targets metadata

phenaproxima:
Partially addresses #60 (for targets metadata).


tedbow:
Sorry it was probably me that suggested we do this. but now I think we should just add getters as we need them for functionality in the PRs for the Updater. Otherwise we may end up with getters we don't use or we may end up needing a different getter, for instance maybe we will just need `getTarget($fileName)` instead.

I am in favor of closing this

phenaproxima:
I agree with @tedbow. Let's add what we need, when we need it. Otherwise we're in danger over over-engineering this thing and introducing unnecessary complexity, which will take time and effort to refactor out.

tedbow:
Ok closing

Issue #33, opened 2020-08-10
Add tests for the FileStorage


xjm:
The default file storage lacks test coverage. Add tests for it.

PR #87, opened 2020-11-13
Add unit test for FileStorage

phenaproxima:
Addresses #33.

### DIFF
--- a/tests/Unit/FileStorageTest.php
+++ b/tests/Unit/FileStorageTest.php
@@ -39,6 +39,6 @@ class FileStorageTest extends TestCase
         $this->assertTrue(isset($storage[$filename]));
         $this->assertSame("From hell's heart, I refactor thee!", $storage[$filename]);
         unset($storage[$filename]);
-        $this->assertFileDoesNotExist("$dir/$filename");
+        $this->assertFileNotExists("$dir/$filename");
     }
 }
--- a/tests/Unit/FileStorageTest.php
+++ b/tests/Unit/FileStorageTest.php
@@ -6,7 +6,7 @@ use PHPUnit\Framework\TestCase;
 use Tuf\Client\DurableStorage\FileStorage;
 
 /**
- * @covers \Tuf\Client\DurableStorage\FileStorage
+ * @coversDefaultClass \Tuf\Client\DurableStorage\FileStorage
  */
 class FileStorageTest extends TestCase
 {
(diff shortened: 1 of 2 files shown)


tests/Unit/FileStorageTest.php
```diff
@@ -0,0 +1,44 @@
...
+     * @return void
+     */
+    public function testStorage(): void
+    {
+        $dir = sys_get_temp_dir();
+        $storage = new FileStorage($dir);
+        $filename = uniqid();
+        $this->assertFalse(isset($storage[$filename]));
+        $storage[$filename] = "From hell's heart, I refactor thee!";
+        $this->assertFileExists("$dir/$filename");
```
xjm, line 38: This tests `offsetGet()` in that case.

tests/Unit/FileStorageTest.php
```diff
@@ -0,0 +1,44 @@
...
+        $this->assertTrue(isset($storage[$filename]));
+        $this->assertSame("From hell's heart, I refactor thee!", $storage[$filename]);
+        unset($storage[$filename]);
```
xjm, line 41: And this tests `offsetUnset()`.

xjm (commented):
Seems straightforward. I think we could either merely mention online which `\ArrayAccess` method were tested, or just explicitly test them?

Is it worth also testing the protected method?

tests/Unit/FileStorageTest.php
```diff
@@ -0,0 +1,44 @@
...
+     * Tests creating and interacting with files via a FileStorage object.
+     *
+     * @return void
+     */
+    public function testStorage(): void
+    {
+        $dir = sys_get_temp_dir();
+        $storage = new FileStorage($dir);
+        $filename = uniqid();
+        $this->assertFalse(isset($storage[$filename]));
```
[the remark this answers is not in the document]
phenaproxima: I figure it's not worth calling the interface methods directly, since that defeats the point of using ArrayAccess in the first place.

phenaproxima:
I usually do not directly test protected methods, since that tightly couples the test to internal implementation details of the class. But I do want to make sure that paths _through_ the protected methods are tested, which I _believe_ they are with this test.

tedbow:
I just added comments for `@covers` . this looks good

tedbow (approved):
Looks good

PR #88, opened 2020-11-16
Fix normalization 

tedbow:
In #80 we found out that are normalization is not working correctly. We only found this then because we had not hit the bug yet.

If run generate the fixtures or download the latest from https://github.com/php-tuf/php-tuf/suites/1494322505/artifacts/26022146 
Take a look at fixtures/TUFTestFixtureDelegated/tufrepo/metadata/1.targets.json
In this file you can see that in the elements
signed/delegations/keys = {}
signed/delegations/role = []
We were validating targets files yet in main which why we hadn't see this.

So 1 is empty object and 1 is an empty array
Because in `\Tuf\Metadata\MetadataBase::createFromJson()` we are using
`$data = json_decode($json, true);` 
and the 2nd argument here convert all objects into arrays then we re-encode to check signatures this will not match. This is not a problem for not empty object because json_encode() will make them objects(no assoc array in JS I think)

@davidstrauss and looked into just manually making sure all empty objects in the JSON are `{}` but this doesn't work because it messes with our validation in `\Tuf\Metadata\MetadataBase::validateMetaData()` we can't have `keys` sometimes be an array and sometimes be an object.

So the first try to fix this I used `json_decode($json)` which preserves the objects and `stdClass` whether there are empty or not. The problem with this is the `symfony/validator` library does not play nice with `stdClass` where the properties are not defined in code.

So to fix this I created `\Tuf\Metadata\ValidatableClass` this class allows use it with `symfony/validator` because of `implements \ArrayAccess, \Iterator, \Countable`. For instance this allows use to the [Collection](https://symfony.com/doc/current/reference/constraints/Collection.html) constraint which requires

> i.e. an array or an object that implements Traversable and ArrayAccess


davidstrauss:
What if we performed signature validation and then did a second pass to replace stdClass objects -- which are known at that point to be empty by construction -- with empty arrays? Unlike bifurcating the processing into separate validation and usage paths -- which was an approach I opposed in case there are quirks that allow getting data in one but not the other -- performing additional processing on the signature-validated data to make later processing easier seems like it would allow integrating the approach we tested on Friday with the remaining processing chain.

That said, I don't think a validation class is a bad idea; it even has its own advantages if we can check for other schema expectations. Happy to do a full review if I can hear feedback on the approach above as a potentially simpler alternative.

tedbow:
@davidstrauss re

> What if we performed signature validation and then did a second pass to replace stdClass objects
I thought about this but if you look at the [client workflow in the spec ](https://github.com/theupdateframework/specification/blob/v1.0.9/tuf-spec.md#5-detailed-workflows) under **5.4.1. Check against snapshot metadata** you will see that we need to read the metadata for the version _before_ **5.4.2. Check for an arbitrary software attack** where we actually check for the signatures.

I am not why it has to be in this order I suppose it could be a mistake. But also since before 5.4.1 we have to load the file and run validation I don't think there is any risk in accessing this metadata before the signature validation.  

So If we had to do signature validation first this would not allow the steps in this order. I think we should stick to the spec order as much as we can and also try not to architected the system where we are forced into a particular.

davidstrauss:
> I am not why it has to be in this order I suppose it could be a mistake.

It seems broken to do anything but essential processing of the signed data before signature validation. Would you mind checking with upstream to understand the rationale? I'd prefer to validate the signature and then check version numbers; it seems like this would also be necessary to achieve TUF's proposed goal of dropping most or all parsing/canonicalization prior to signature validation.

tedbow:
@davidstrauss I gave some of this a try in #89 
The problem is for root you have actually validate the structure before you can verify the signature because roles and the keys are in the structure. So we still have to have the array version to validate unless we are still going to have `ValidatableClass`

Also verifying the signature requires us to read the signed portion anyways so we are now interacting with before we validated the structure that we are reading. 

one thing I do like about #89 is creating a `SignatureVerifier` class to move all of the signature logic into that class. 

Also passing the `SignatureVerifier` instance to `createFromJson` so that it is validated in factory. This would force signature to happen before we did any other logic. You then could not create an instance of a `MetadataBase` without it being verified.

We would still have to make sure the order of the steps didn't matter above.

src/Metadata/ValidatableClass.php
```diff
@@ -0,0 +1,25 @@
...
+
+use function DeepCopy\deep_copy;
+
+/**
+ * Defines a class that can used with Symfony validator constraints.
+ *
+ * Symfony validator constraints do not work with classes that have dynamic
+ * public properties.
+ */
+final class ValidatableClass extends \ArrayObject
```
tedbow, line 13: This class is really simple now. We could use ArrayObject directly except that we want to make sure that `offsetGet()` always returns clones so that the metadata is not altered after it is set.

pwolanin:
seems like a reasonable direction instead of special-casing properties

composer.json
```diff
@@ -14,7 +14,8 @@
         "ext-json": "*",
         "paragonie/sodium_compat": "^1.13",
         "symfony/validator": "^5.1",
-        "guzzlehttp/guzzle": "^7.2"
+        "guzzlehttp/guzzle": "^7.2",
+        "myclabs/deep-copy": "^1.10"
```
tedbow, line 18: Figured this was better than writing our own it. It is already required by phpunit but this makes it non-dev requirement.

tests/Metadata/MetaDataBaseTest.php
```diff
@@ -411,4 +412,22 @@ protected function getFixtureNestedArrayFirstKey(string $fixtureName, array $nes
...
+     * @return void
+     *
+     * @dataProvider providerValidMetaData
+     */
+    public function testNormalization(string $validJson) : void
+    {
+        $contents = $this->localRepo[$validJson];
+        $json = json_decode($contents);
+        $metaData = static::callCreateFromJson($contents);
+        $this->assertEquals(json_encode($json->signed), JsonNormalizer::asNormalizedJson($metaData->getSigned()));
```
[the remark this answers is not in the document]
tedbow: Added a unit test for JsonNormalizer

src/Metadata/ValidatableClass.php
tedbow: So the reason we needed this was because `\Tuf\KeyDB::createFromRootMetadata` was adding info to the key info. This didn't matter before but because this is now an object it is returned by reference. So the idea here is it that our metadata should not be change when a getter is used. 

but I think it is better to handle this on the getters for the metadata. So I am going to add `deep_copy()` there. This will allow remove this class completely and just `ArrayObject` everywhere we were using this.

davidstrauss:
> We would still have to make sure the order of the steps didn't matter above.

Any findings here?

tedbow:
> > We would still have to make sure the order of the steps didn't matter above.
> 
> Any findings here?

@davidstrauss no I haven't had a chance to ask yet. But I don't think this it is necessary for this PR any more now that we no longer need `ValidatableClass `  and now think the order is unrelated to this PR.  Also the since this PR is off `main` and not #80 we don't actually have the step where we would call `getVersion()` before we verified the signature. So far we always calling calling `\Tuf\Client\Updater::checkSignatures` directly after `\Tuf\Metadata\MetadataBase::createFromJson()` this PR doesn't change that.

See my comment above 

> Also verifying the signature requires us to read the signed portion anyways so we are now interacting with before we  validated the structure that we are reading.

(more details above) 

#89 explores the idea of moving in signature verification into `\Tuf\Metadata\MetadataBase::createFromJson()` which would mean it would impossible to use any of our metadata getter methods until after signature verification. I could move onto that after this PR. Even if that PR doesn't work we can still see in #80 if we there is any reason we can't verify the signature of the targets file before checking the version.

src/Metadata/RootMetadata.php
```diff
@@ -59,7 +60,7 @@ protected static function getSignedCollectionOptions(): array
      */
     public function getRoles()
     {
-        return $this->getSigned()['roles'];
+        return deep_copy($this->getSigned()['roles']);
```
tedbow, line 63: In any of the getters where we are already calling `getSigned()` we don't need to call `deep_copy()` again because `getSigned()` already calls this and gives a us a clone

davidstrauss (approved):
Very nice work. The type specifications are probably more correct, anyway (to specify the broadest type accepted instead of the narrowest).

trishankatdatadog:
For _loading_ local root, yes, you need to be able to parse the signed metadata w/o first checking signatures (because you need to parse and find the root keys first in order to verify self-signed signatures).

For the snapshot and targets metadata, I agree it's not a great idea that we're checking version and/or hashes before checking signatures. Would one of you gents be so kind as to file an issue with the upstream spec?

trishankatdatadog:
For _loading_ local root, yes, you need to be able to parse the signed metadata w/o first checking signatures (because you need to parse and find the root keys first in order to verify self-signed signatures).

For the snapshot and targets metadata, I believe the spec recommends checking only the hashes, right after download, before checking signatures. This is fine, since the hashes are obtained from other, previously verified sources (timestamp and snapshot for snapshot and targets respectively). The spec does recommend checking version after checking signatures. Does this help to clarify?

tedbow:
@trishankatdatadog yep that helps!

PR #89, opened 2020-11-17
Ensure metadata factory methods return signature Verified metadata

tedbow:
Currently we have a few spots that do something like

`$newTimestampData = TimestampMetadata::createFromJson($newTimestampContents);
  $this->checkSignatures($newTimestampData);`

It's good to `checkSignatures()` before we interact more with the data. If we moved the verification into `createFromJson()` then we could ensure that they return verified metadata.


davidstrauss:
I'm a little confused. The comment here suggests this PR is an alternative to #88 (now merged), but most of the diff seems to be refactoring. Is there a better high-level description of what I should be looking for in this review other than commenting on the design changes in the refactoring?

tedbow:
@davidstrauss yep updated the comment

davidstrauss (approved):
These changes look good to me, though I'm not intimately familiar with the current architecture of the PHP. I like what appears to be dependency injection for the signature verification.

tedbow:
Not sure if this need any more. I do like the idea of having a `SignatureVerifier`

tedbow:
Closing this in favor of #198

PR #90, opened 2020-12-03
\Tuf\Tests\Client\TestRepo::fetchFile interface fix

tedbow:
\Tuf\Tests\Client\TestRepo::fetchFile incorrectly returned a false when no file when interface says to throw exception

`\Tuf\Client\Updater::updateRoot()` was depending on the this broken behavior so a new method `\Tuf\Client\RepoFileFetcherInterface::fetchFileIfExists()` was added to make this cleaner.

This fix was in #80  but it is really unrelated just discovered there

### DIFF
--- a/src/Client/RepoFileFetcherInterface.php
+++ b/src/Client/RepoFileFetcherInterface.php
@@ -2,6 +2,8 @@
 
 namespace Tuf\Client;
 
+use Tuf\Exception\RepoFileNotFound;
+
 /**
  * Defines an interface for fetching repo files.
  */
@@ -26,5 +28,18 @@ interface RepoFileFetcherInterface
      * @throws \Tuf\Exception\DownloadSizeException
      *   Thrown if the file exceeds $maxBytes in size.
      */
-    public function fetchFile(string $fileName, int $maxBytes);
+    public function fetchFile(string $fileName, int $maxBytes):string;
+
+    /**
+     * Gets a file if it exists if it exists in the remote repo.
+     *
+     * @param string $fileName
+     *   The file name to fetch.
+     * @param integer $maxBytes
+     *   The maximum number of bytes to download.
+     *
+     * @return string|null
+     *   The contents of the file or null if it does not exist.
+     */
+    public function fetchFileIfExists(string $fileName, int $maxBytes):?string;
 }
--- a/src/Client/RepoFileFetcherInterface.php
+++ b/src/Client/RepoFileFetcherInterface.php
@@ -31,7 +31,7 @@ interface RepoFileFetcherInterface
     public function fetchFile(string $fileName, int $maxBytes):string;
 
     /**
-     * Gets a file if it exists if it exists in the remote repo.
+     * Gets a file if it exists in the remote repo.
      *
      * @param string $fileName
      *   The file name to fetch.
--- a/tests/Client/TestRepo.php
+++ b/tests/Client/TestRepo.php
@@ -57,7 +57,7 @@ class TestRepo implements RepoFileFetcherInterface
             }
             return $contents;
         } catch (\Exception $exception) {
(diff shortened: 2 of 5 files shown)


phenaproxima (approved):

src/Client/RepoFileFetcherInterface.php
```diff
@@ -26,5 +28,18 @@ interface RepoFileFetcherInterface
      * @throws \Tuf\Exception\DownloadSizeException
      *   Thrown if the file exceeds $maxBytes in size.
      */
-    public function fetchFile(string $fileName, int $maxBytes);
+    public function fetchFile(string $fileName, int $maxBytes):string;
+
+    /**
+     * Gets a file if it exists if it exists in the remote repo.
```
phenaproxima, line 34: "if it exists" is repeated here

tests/Client/TestRepo.php
```diff
@@ -57,7 +57,19 @@ public function fetchFile(string $fileName, int $maxBytes)
             }
             return $contents;
         } catch (\Exception $exception) {
-            return false;
+            throw new RepoFileNotFound("File $fileName not found.");
```
phenaproxima, line 60: The RepoFileNotFound exception should wrap the previous exception, to make debugging easier.

tests/Unit/GuzzleFileFetcherTest.php
```diff
@@ -85,31 +100,55 @@ public function errorProvider(): array
      *
      * @return void
      *
-     * @dataProvider errorProvider
+     * @dataProvider providerFetchFileError
      *
      * @covers ::fetchFile
      */
-    public function testError(int $statusCode, string $exceptionClass, ?int $exceptionCode = null, ?int $maxBytes = null): void
+    public function testfetchFileError(int $statusCode, string $exceptionClass, ?int $exceptionCode = null, ?int $maxBytes = null): void
```
phenaproxima, line 107: Nit, should be testFetchFileError (first 'f' is not capitalized)

tests/Unit/GuzzleFileFetcherTest.php
```diff
@@ -127,4 +166,25 @@ public function testCreateFromUri(): void
...
+     * @param integer|null $exceptionCode
+     *   (optional) The expected exception code. Defaults to the status code.
+     *
+     * @return GuzzleFileFetcher
+     *   The fetcher that wll producer an exception.
+     */
+    protected function getFetcherWithError(int $statusCode, string $exceptionClass, ?int $exceptionCode): GuzzleFileFetcher
+    {
+        $this->mockHandler->append(new Response($statusCode, [], $this->testContent));
+        $this->expectException($exceptionClass);
```
phenaproxima, line 186: Not sure how I feel about calling $this->expectException() and $this->expectExceptionCode() in a helper method; it obfuscates the expectations of the test. Not a big deal, but feels a bit "smelly" to me.

phenaproxima (changes requested):

src/Client/RepoFileFetcherInterface.php
tedbow: fixed

tests/Client/TestRepo.php
tedbow: fixed

tests/Unit/GuzzleFileFetcherTest.php
tedbow: fixed

tests/Unit/GuzzleFileFetcherTest.php
tedbow: yeah. I just removed this helper function. Agreed these are better in caller methods and then the new method really isn't necessary

phenaproxima (approved):

PR #91, opened 2020-12-08
Add testing for file not file exception

tedbow:
#90 fixed the TestRepo class to thrown exception instead of returning FALSE on a file not found. This was originally in #80 but we split it off because it was unrelated. But we forgot to pull out all the test coverage for this fix that was in #80 

This is the test coverage. It will also make #80 even smaller

### DIFF
--- a/tests/Client/TestRepo.php
+++ b/tests/Client/TestRepo.php
@@ -5,26 +5,21 @@ namespace Tuf\Tests\Client;
 use Tuf\Client\RepoFileFetcherInterface;
 use Tuf\Exception\RepoFileNotFound;
 use Tuf\JsonNormalizer;
+use Tuf\Tests\TestHelpers\UtilsTrait;
 
 /**
  * Defines an implementation of RepoFileFetcherInterface to use with test fixtures.
  */
 class TestRepo implements RepoFileFetcherInterface
 {
+    use UtilsTrait;
 
     /**
-     * The fixtures set to use.
-     *
-     * @var string
-     */
-    protected $fixturesSet;
-
-    /**
-     * File names for files that should fail signature checks.
+     * An array of repo file contents keyed by file name.
      *
      * @var string[]
      */
-    protected $failSignatureFiles = [];
+    private $repoFilesContents = [];
 
     /**
      * TestRepo constructor.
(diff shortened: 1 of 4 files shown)


tests/Client/TestRepo.php
```diff
@@ -34,31 +29,23 @@ class TestRepo implements RepoFileFetcherInterface
      */
     public function __construct(string $fixturesSet)
     {
-        $this->fixturesSet = $fixturesSet;
+        // Store all the repo files locally so they can be easily altered.
+        // @see self::setRepoFileNestedValue()
+        $repoFiles = glob(static::getFixturesRealPath($fixturesSet, '/tufrepo/metadata') . '/*.json');
+        foreach ($repoFiles as $repoFile) {
+            $this->repoFilesContents[basename($repoFile)] = file_get_contents($repoFile);
```
phenaproxima, line 36: I feel like this is going to potentially chew up a lot of memory. Is this really the right way to go?

tests/Client/TestRepo.php
```diff
@@ -74,15 +61,34 @@ public function fetchFileIfExists(string $fileName, int $maxBytes):?string
...
+     * @param string $fileName
+     *   The name of the file to change.
+     * @param array $keys
+     *   The nested array keys of the item.
+     * @param mixed $newValue
+     *   The new value to set.
+     *
+     * @return void
+     */
+    public function setRepoFileNestedValue(string $fileName, array $keys = ['signed', 'extra_test_value'], $newValue = 'new value'): void
```
phenaproxima, line 75: Why do $keys and $newValue have defaults? That seems strange.

tests/TestHelpers/UtilsTrait.php
```diff
@@ -64,4 +66,26 @@ protected static function getKeyedArray(array $providedData, int $useArgumentNum
...
+     * @param array $keys
+     *   Ordered keys to the value to set.
+     * @param array $data
+     *   The array to modify.
+     * @param mixed $newValue
+     *   The new value to set.
+     *
+     * @return void
+     */
+    protected function nestedChange(array $keys, array &$data, $newValue) : void
```
phenaproxima, line 82: Not in scope, but nestedChange() is such a weird name for this method. Maybe setNested() or just set() would be preferable?

phenaproxima (changes requested):

tests/Client/TestRepo.php
tedbow: It will use more memory but this is only for testing. Right this is at most 26 files per test. when I run `composer run test --profile` I don't see any different in memory usage, both say `8.6MiB` from `main`

The fact that we have the contents in memory also means we can really simplify `fetchFile()` and makes it super easy for us to add the new `removeRepoFile()`. 

I would much rather for code simplicity then worry about memory usage in tests. In main and this PR I can run the test suite with minimum `memory_limit = 20M`. So don't think we are close to a problem

tests/TestHelpers/UtilsTrait.php
tedbow: Yeah this is just moving an existing method. but I agree it could be named better

tests/Client/TestRepo.php
tedbow: So we can used the method to set a nested value when we don't actually care what the value is just that contents have changed. Right now this is the only use case for this method but it is being pull from #80 which hopefully will be in soon. In that case we do have case we want to change a specific key.

PR #92, opened 2020-12-10
 Assert start versions of the client repo metadata in all updater tests

tedbow:
In `providerRefreshRepository()` we are providing expected start versions of the client repo *and* the test fixture set as arguments. But the start versions of the client repo are always the same for the fixture set so we just need to provide fixture set.

Also else where we only assert the start version of the root metadata and not others.


phenaproxima (commented):
This looks okay on its face, but I don't know how qualified I am to sign off on it, since I don't really *understand* what it's doing.

tests/Client/UpdaterTest.php
```diff
@@ -254,6 +273,7 @@ public function testFileNotFoundExceptions(string $fixturesSet, string $fileName
         // side-effects.
         $this->localRepo = $this->memoryStorageFromFixture($fixturesSet, 'tufclient/tufrepo/metadata/current');
         $this->testRepo = new TestRepo($fixturesSet);
+        $this->assertClientRepoVersions($this->getFixtureClientStartVersions($fixturesSet));
```
phenaproxima, line 276: Shouldn't this be using static::getFixtureClientStartVersions()?

phenaproxima (approved):
@tedbow walked me through this on Zoom and I think it makes a lot of sense.

I still find the tests as a whole kinda hard to understand, mostly due to method naming and some code organization decisions. All of that is outside the scope of this PR, but should probably be refactored once we have implemented the spec.

tests/Client/UpdaterTest.php
tedbow: yep fixed

PR #94, opened 2020-12-10
Complete spec 5.3.1 by checking file hash from timestamp

tedbow:
In the client logic of the spec https://github.com/theupdateframework/specification/blob/v1.0.9/tuf-spec.md#5-detailed-workflows

Step **5.3.1. Check against timestamp metadata** says we should check the new snapshot metadata version and hash against the information in the snapshot file. Currently we are only checking the version.


tests/Client/UpdaterTest.php
```diff
@@ -188,69 +189,112 @@ protected function assertClientRepoVersions(array $expectedVersions): void
...
             $updater->refresh();
-        } catch (SignatureThresholdExpception $exception) {
-            // Confirm the root was updated to version 4 but not to 5 because
-            // 5.root.json should throw an exception.
-            $this->assertSame(
-                $expectedRootVersion,
-                RootMetadata::createFromJson($this->localRepo['root.json'])->getVersion()
-            );
-            $this->assertSame($expectionMessage, $exception->getMessage());
+        } catch (\Exception $exception) {
```
phenaproxima, line 220: Maybe we should use `TufException` here, since we probably shouldn't be throwing any _other_ kind of exception from the updater?

tests/Client/UpdaterTest.php
```diff
@@ -188,69 +189,112 @@ protected function assertClientRepoVersions(array $expectedVersions): void
...
+            $this->assertEquals($exception, $expectionException);
+            $this->assertClientRepoVersions($expectedUpdatedVersions);
             return;
         }
-        $this->fail('No SignatureThresholdExpception thrown');
+        $this->fail('No exception thrown. Expected: ' . get_class($expectionException));
```
phenaproxima, line 225: Can't we use `$this->expectException()` instead?

phenaproxima (changes requested):

tests/Client/UpdaterTest.php
tedbow: I don't think so because we want to check `$this->assertClientRepoVersions($expectedUpdatedVersions);` after the exception is thrown. 

For instance for the root metadata object is updated because the new one is valid that change should persist even if we get an exception on the snapshot. But the snapshot should not persist. So need to check the version after the exception

tests/Client/UpdaterTest.php
```diff
@@ -188,69 +189,112 @@ protected function assertClientRepoVersions(array $expectedVersions): void
...
-    public function testSignatureError(string $fileToFail, int $expectedRootVersion, string $expectionMessage) : void
+    public function testRefreshException(string $fileToChange, array $keys, $newValue, \Exception $expectionException, array $expectedUpdatedVersions): void
     {
         // Use the memory storage used so tests can write without permanent
         // side-effects.
         $this->localRepo = $this->memoryStorageFromFixture('TUFTestFixtureDelegated', 'tufclient/tufrepo/metadata/current');
         $this->testRepo = new TestRepo('TUFTestFixtureDelegated');
-        $this->assertClientRepoVersions(static::getFixtureClientStartVersions('TUFTestFixtureDelegated'));
-        $this->testRepo->setRepoFileNestedValue($fileToFail);
+        $this->assertClientRepoVersions($this->getFixtureClientStartVersions('TUFTestFixtureDelegated'));
```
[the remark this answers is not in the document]
tedbow: fixed

tests/Client/UpdaterTest.php
tedbow: fixed

phenaproxima (approved):
Honestly, I like it. I also like the fact that we now have the ability for metadata objects to verify other metadata objects directly. Since that verification is idempotent and doesn't normally need anything apart from the incoming metadata object, I find it very clean and simple to read, use, and understand. It's also in accordance, IMHO, with the way the spec works.

PR #95, opened 2020-12-11
KeyDB::getKey() will return metadata with unused sha512 'keyid' array key for both 256 and 512 key ids.

tedbow:
Work @bircher and @ZeiP at DrupalCon Europe 2020 contribution day.

Originally we were going reroll #49 but we found this bug in the process. 

It was probably caused when we switched from using array to ArrayObjects because they are passed by reference

### DIFF
--- a/src/KeyDB.php
+++ b/src/KeyDB.php
@@ -46,15 +46,8 @@ class KeyDB
         $db = new self();
 
         foreach ($rootMetadata->getKeys() as $keyMeta) {
-            if (! in_array($keyMeta['keytype'], self::getSupportedKeyTypes(), true)) {
-                // @todo Convert this to a log line as per Python.
-                throw new \Exception("Root metadata file contains an unsupported key type: \"${keyMeta['keytype']}\"");
-            }
-            // One key ID for each $keyMeta['keyid_hash_algorithms'].
-            $computedKeyIds = self::computeKeyIds($keyMeta);
-            foreach ($computedKeyIds as $keyId) {
-                $db->addKey($keyId, $keyMeta);
-            }
+            $db->addKey($keyMeta);
+
         }
 
         return $db;
@@ -138,9 +131,17 @@ class KeyDB
      *
      * @todo https://github.com/php-tuf/php-tuf/issues/56
      */
-    private function addKey(string $keyId, \ArrayAccess $keyMeta)
+    private function addKey(\ArrayAccess $keyMeta)
     {
-        $this->keys[$keyId] = $keyMeta;
+        if (! in_array($keyMeta['keytype'], self::getSupportedKeyTypes(), true)) {
+            // @todo Convert this to a log line as per Python.
+            throw new \Exception("Root metadata file contains an unsupported key type: \"${keyMeta['keytype']}\"");
+        }
+        // One key ID for each $keyMeta['keyid_hash_algorithms'].
+        $computedKeyIds = self::computeKeyIds($keyMeta);
+        foreach ($computedKeyIds as $keyId) {
+            $this->keys[$keyId] = $keyMeta;
+        }
     }
 
     /**
--- a/src/KeyDB.php
+++ b/src/KeyDB.php
@@ -4,6 +4,7 @@
 namespace Tuf;
 
 use Tuf\Metadata\RootMetadata;
+use function DeepCopy\deep_copy;
 
 /**
  * Represent a collection of keys and their organization.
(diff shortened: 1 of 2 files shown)


src/KeyDB.php
```diff
@@ -52,8 +53,7 @@ public static function createFromRootMetadata(RootMetadata $rootMetadata)
             // One key ID for each $keyMeta['keyid_hash_algorithms'].
             $computedKeyIds = self::computeKeyIds($keyMeta);
             foreach ($computedKeyIds as $keyId) {
-                $keyMeta['keyid'] = $keyId;
-                $db->addKey($keyMeta);
```
tedbow, line 56: The only reason why `$keyMeta['id'] was set was so that `addKey()` could use it when setting `$this->keys[$keyId]`. Since $keyMeta is not an object means that all the previous items set using `$this->keys` actual refer to the same object. So the object will have the `keyId` of the last element of `$computedKeyIds`

We don't need to set `keyId` at all and can move the call to `computeKeyIds()` inside `addKey()`. That the process of adding items to `$this->keys` we be encapsulated inside `addKey()`. This is will make added the keys from `targets.json` simpler.

src/KeyDB.php
```diff
@@ -156,11 +156,11 @@ public function addKey(\ArrayAccess $keyMeta)
      *
      * @see https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#4-document-formats
      */
-    public function getKey(string $keyId)
+    public function getKey(string $keyId):\ArrayObject
```
phenaproxima, line 159: IMHO the return type hint should be `\ArrayAccess`, since that's an interface, as opposed to a concrete class.

tests/KeyDBTest.php
```diff
@@ -4,53 +4,51 @@
...
+                'sha512',
             ],
+            'keytype' => 'ed25519',
+            'keyval' => new \ArrayObject(['public' => '6400d770c7c1bce4b3d59ce0079ed686e843b6500bbea77d869a1ae7df4565a1']),
+            'scheme' => 'ed25519',
         ];
+        $expectedKeyDataObject = new \ArrayObject($expectedKeyData);
+        $key256 = $keyDb->getKey($keyId256);
+        $key512 = $keyDb->getKey($keyId512);
+        self::assertEquals($expectedKeyDataObject, $key256);
```
phenaproxima, line 46: Are we sure that PHPUnit handles ArrayObject as we expect it to here? Maybe we might want to do something like `self::assertSame($expectedKeyDataObject->getArrayCopy(), $key256->getArrayCopy())` instead?

phenaproxima (changes requested):

src/KeyDB.php
```diff
@@ -136,9 +128,17 @@ public function __construct()
      *
      * @todo https://github.com/php-tuf/php-tuf/issues/56
      */
-    public function addKey(\ArrayAccess $keyMeta)
+    private function addKeyMetadata(\ArrayAccess $keyMeta)
     {
-        $this->keys[$keyMeta['keyid']] = $keyMeta;
+        if (! in_array($keyMeta['keytype'], self::getSupportedKeyTypes(), true)) {
+            // @todo Convert this to a log line as per Python.
+            throw new \Exception("Root metadata file contains an unsupported key type: \"${keyMeta['keytype']}\"");
```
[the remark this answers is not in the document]
tedbow: This was just an existing line that was moved inside this method.  I would rather wait and see if this exception is removed all together as per the `@todo` comment above

src/KeyDB.php
tedbow: We have other case in the project that are using `\ArrayObject` but none that are using `\ArrayAccess`.  So I would rather keep in the same.  Also `\ArrayObject` implements 3 other interfaces so it actually is more descriptive. We know will be an `\ArrayObject`

tests/KeyDBTest.php
tedbow: Hmmm, good point but your suggestion won't work because the nested \ArrayObject is not the exact same instance.

I am going to use `json_encode()` instead of `getArrayCopy()`.

src/KeyDB.php
```diff
@@ -136,9 +128,17 @@ public function __construct()
      *
      * @todo https://github.com/php-tuf/php-tuf/issues/56
      */
-    public function addKey(\ArrayAccess $keyMeta)
+    private function addKeyMetadata(\ArrayAccess $keyMeta)
```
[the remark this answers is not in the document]
tedbow: undid the change

phenaproxima (approved):

Issue #97, opened 2020-12-15
[META] Ensure test metadata fixtures cover all possible elements in specification


tedbow:
Currently our test fixtures do not cover all elements or element value that may be  in the [specification client workflow](https://github.com/theupdateframework/specification/blob/v1.0.9/tuf-spec.md#5-detailed-workflows). We should either handle all these value or throw explicit exceptions if there are elements we don't handle. It may be easiest to throw exceptions in `\Tuf\Metadata\MetadataBase::validateMetaData()` or override this is extending classes.

Elements we currently don't support

1. `consistent_snapshot` == FALSE
2. `hashes` in `snapshots.json` for `targets.json` (the code supports it but the test fixtures don't have it)
3. `length` in `snapshots.json` for `targets.json`

davidstrauss:
Adding fixtures without consistent snapshots is easy. I've added #98 to capture this.

I'm not sure how much control we have in Python of the hash/length data in snapshots for targets; the ability to control that certainly isn't exposed in the Python API in an obvious way. I think we may need to reach upstream to handle these in a clean and flexible way.

tedbow:
I changed the name of this PR. I don't think we need to have all optional elements in the spec in our tests fixtures.  I just think our tests need to account for them.

Accounting for the optional fields may just mean that we make sure there is an error if we find an optional field that we don't support we throw an error. I created #99 for this

My opinion is that we should perhaps explicitly declare that we support metadata files as produced by the Python library. Otherwise it seems like we would have to make test fixtures from another TUF server library. To me this sounds like taking more work than we need to right now.

Looking at [version of the specification](https://github.com/theupdateframework/specification/blob/v1.0.9/tuf-spec.md) we are working off. Only "MAY"(capitalized which has a special meaning for implementers) once in the document

> The snapshot.json file is signed by the snapshot role. It MUST list the version numbers of the top-level targets metadata and all delegated targets metadata. It MAY also list their lengths and file hashes.

So this might be the only case there is optional field in the file formats. But it would good have general decision around this.

PR #99, opened 2020-12-15
Throw metadata validation exception on Unsupported fields

tedbow:
This may not necessary but it seems like there are some optional fields that the Python lib might not support creating.

### DIFF
--- a/src/Constraints/Collection.php
+++ b/src/Constraints/Collection.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace Tuf\Constraints;
+
+use Symfony\Component\Validator\Constraints\Collection as SymfonyCollection;
+
+/**
+ * Custom constraint that extends Symfony's Collection constraint to add 'excludedFields'.
+ */
+class Collection extends SymfonyCollection
+{
+
+    /**
+     * @var array
+     */
+    public $unsupportedFields = [];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function __construct($options = null)
+    {
+        if (isset($options['excludedFields']) && !isset($options['fields'])) {
+            throw new \InvalidArgumentException('If "excludedFields" is set "fields" must also be set.');
+        }
+        parent::__construct($options);
+    }
+}
--- a/src/Constraints/Collection.php
+++ b/src/Constraints/Collection.php
@@ -14,15 +14,4 @@ class Collection extends SymfonyCollection
      * @var array
      */
     public $unsupportedFields = [];
-
-    /**
-     * {@inheritdoc}
-     */
-    public function __construct($options = null)
-    {
-        if (isset($options['excludedFields']) && !isset($options['fields'])) {
-            throw new \InvalidArgumentException('If "excludedFields" is set "fields" must also be set.');
-        }
-        parent::__construct($options);
-    }
 }
--- a/src/Constraints/CollectionValidator.php
+++ b/src/Constraints/CollectionValidator.php
@@ -14,20 +14,17 @@ class CollectionValidator extends SymfonyCollectionValidator
      */
     public function validate($value, Constraint $constraint)
     {
-        if (!empty($constraint->unsupportedFields)) {
-            foreach ($constraint->unsupportedFields as $unsupportedField) {
-                $existsInArray = \is_array($value) && \array_key_exists($unsupportedField, $value);
-                $existsInArrayAccess = $value instanceof \ArrayAccess && $value->offsetExists($unsupportedField);
-                if ($existsInArray || $existsInArrayAccess) {
-                    $this->context->buildViolation('This field is not supported.')
-                      ->atPath('['.$unsupportedField.']')
-                      ->setInvalidValue(null)
-                      ->setCode(Collection::MISSING_FIELD_ERROR)
-                      ->addViolation();
-                }
+        foreach ($constraint->unsupportedFields as $unsupportedField) {
+            $existsInArray = \is_array($value) && \array_key_exists($unsupportedField, $value);
+            $existsInArrayAccess = $value instanceof \ArrayAccess && $value->offsetExists($unsupportedField);
+            if ($existsInArray || $existsInArrayAccess) {
+                $this->context->buildViolation('This field is not supported.')
+                  ->atPath("[$unsupportedField]")
+                  ->setInvalidValue(null)
+                  ->setCode(Collection::MISSING_FIELD_ERROR)
+                  ->addViolation();
             }
         }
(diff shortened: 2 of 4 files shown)


src/Constraints/Collection.php
```diff
@@ -0,0 +1,28 @@
...
+     */
+    public $unsupportedFields = [];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function __construct($options = null)
+    {
+        if (isset($options['excludedFields']) && !isset($options['fields'])) {
+            throw new \InvalidArgumentException('If "excludedFields" is set "fields" must also be set.');
```
phenaproxima, line 24: Feel free to ignore, but IMHO LogicException is the more semantically proper choice here.

src/Constraints/Collection.php
```diff
@@ -0,0 +1,28 @@
...
+     * @var array
+     */
+    public $unsupportedFields = [];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function __construct($options = null)
+    {
+        if (isset($options['excludedFields']) && !isset($options['fields'])) {
```
phenaproxima, line 23: Is it okay if $options['fields'] is an empty array, or other falsy value? If not, we might want to use `empty()` instead.

src/Constraints/CollectionValidator.php
```diff
@@ -0,0 +1,33 @@
...
+     */
+    public function validate($value, Constraint $constraint)
+    {
+        if (!empty($constraint->unsupportedFields)) {
+            foreach ($constraint->unsupportedFields as $unsupportedField) {
+                $existsInArray = \is_array($value) && \array_key_exists($unsupportedField, $value);
+                $existsInArrayAccess = $value instanceof \ArrayAccess && $value->offsetExists($unsupportedField);
+                if ($existsInArray || $existsInArrayAccess) {
+                    $this->context->buildViolation('This field is not supported.')
+                      ->atPath('['.$unsupportedField.']')
```
phenaproxima, line 23: Is this supposed to be using single quotes? It looks like it should be double-quoted so that $unsupportedField will be replaced...

src/Constraints/CollectionValidator.php
```diff
@@ -0,0 +1,33 @@
...
+
+class CollectionValidator extends SymfonyCollectionValidator
+{
+
+    /**
+     * {@inheritdoc}
+     */
+    public function validate($value, Constraint $constraint)
+    {
+        if (!empty($constraint->unsupportedFields)) {
```
phenaproxima, line 17: I'm not sure we need this enclosing `if` statement, since $constraint->unsupportedFields defaults to an empty array.

src/Metadata/SnapshotMetadata.php
```diff
@@ -29,9 +28,13 @@ protected static function getSignedCollectionOptions(): array
             new Count(['min' => 1]),
             new All([
                 new Collection(
-                    static::getVersionConstraints() +
                     [
-                        'length' => new Optional([new Type('integer')]),
+                        'fields' => static::getVersionConstraints(),
+                        // These fields are mentioned in the specification as optional but the Python library does not
+                        // adding these fields. Since we use the Python library for our fixtures we cannot create test
```
phenaproxima, line 34: s/adding/add

src/Metadata/SnapshotMetadata.php
```diff
@@ -29,9 +28,13 @@ protected static function getSignedCollectionOptions(): array
...
+                        // fixtures that have these fields specified.
+                        'unsupportedFields' => ['length', 'hashes'],
```
phenaproxima, line 36: So...the Python library doesn't add them, but this comment implies that they *are* part of the spec. Do we really want to mark them as unsupported? Maybe we should at least have a TODO here?

tests/Metadata/SnapshotMetadataTest.php
```diff
@@ -50,15 +51,40 @@ public function providerValidField() : array
...
+    public function testUnsupportedFields(array $unsupportedField, $fieldValue):void
+    {
+        $metadata = json_decode($this->localRepo[$this->validJson], true);
+        $this->nestedChange($unsupportedField, $metadata, $fieldValue);
+        $fieldName = array_pop($unsupportedField);
+        $expectedMessage = preg_quote("Object(ArrayObject)[signed][meta][targets.json][$fieldName]", '/');
+        $expectedMessage .= ".*This field is not supported.";
+        $this->expectException(MetadataException::class);
+        $this->expectExceptionMessageMatches("/$expectedMessage/s");
+        static::callCreateFromJson(json_encode($metadata));
```
phenaproxima, line 74: It'd be nice if MetadataBase had a `createFromArray()` method, so that we can pass it stuff like this without needing to re-encode it as JSON. Not a big deal, though!

src/Constraints/Collection.php
tedbow: I actually removed the overridden constructor completely. I don't think we need this check here. It could be the case where would want to use this Collection class without specifying any `fields` in that case you don't care which fields **are** declared just that some fields should not be.

src/Constraints/CollectionValidator.php
tedbow: Removing

src/Constraints/CollectionValidator.php
tedbow: Yep. I had just copied this from the [symfony lib](https://github.com/symfony/validator/blob/5.x/Constraints/CollectionValidator.php#L61). Fixing

src/Metadata/SnapshotMetadata.php
tedbow: fixed

src/Metadata/SnapshotMetadata.php
tedbow: They are optional in the spec 

> The snapshot.json file is signed by the snapshot role. It MUST list the version numbers of the top-level targets metadata and all delegated targets metadata. It MAY also list their lengths and file hashes.

It think if the Python library doesn't add optional fields we shouldn't support them because we don't have way to make test fixtures for them. But see the ongoing discussion here #97

tests/Metadata/SnapshotMetadataTest.php
tedbow: That would need to be public method which I don't think we should add just for testing purposes. But maybe we could add `createMetadataFromArray()` to `\Tuf\Tests\Metadata\MetaDataBaseTest` though in another PR

phenaproxima (approved):

PR #76, opened 2020-11-05
Create test for attacks with new fixtures

tedbow:

PR #101, opened 2020-12-18
Create a rollback attack repo and test

tedbow:
This will replace the outdated #76 

This will just create a Rollback attack fixture.

### DIFF
--- a/generate_fixtures.py
+++ b/generate_fixtures.py
@@ -129,6 +129,19 @@ class TUFTestFixtureSimple(TUFTestFixtureBase):
         self.write_and_publish_repository(export_client=True)
 
 
+class TUFTestFixtureAttackRollback(TUFTestFixtureSimple):
+    def __init__(self):
+        super().__init__()
+        backup_dir = self.tufrepo_dir + "_backup"
+        shutil.copytree(self.tufrepo_dir, backup_dir, dirs_exist_ok=True)
+        self.write_and_add_target('testtarget2.txt')
+        self.write_and_publish_repository(export_client=True)
+        shutil.rmtree(self.tufrepo_dir + '/')
+        # Reset the client to previous state to simulate a rollback attack.
+        shutil.copytree(backup_dir, self.tufrepo_dir, dirs_exist_ok=True)
+        shutil.rmtree(backup_dir + '/')
+
+
 class TUFTestFixtureDelegated(TUFTestFixtureSimple):
     def __init__(self):
         super().__init__()
@@ -177,6 +190,7 @@ class TUFTestFixtureDelegated(TUFTestFixtureSimple):
 def generate_fixtures():
     TUFTestFixtureSimple()
     TUFTestFixtureDelegated()
+    TUFTestFixtureAttackRollback()
 
 
 generate_fixtures()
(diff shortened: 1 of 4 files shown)


generate_fixtures.py
```diff
@@ -129,6 +129,19 @@ def __init__(self):
         self.write_and_publish_repository(export_client=True)
 
 
+class TUFTestFixtureAttackRollback(TUFTestFixtureSimple):
```
phenaproxima, line 132: My Python-fu is not strong enough to offer serious review on this, maybe @davidstrauss should glance at it? On its face, it looks okay to me.

tests/Client/UpdaterTest.php
```diff
@@ -6,6 +6,7 @@
 use PHPUnit\Framework\TestCase;
 use Tuf\Client\Updater;
 use Tuf\Exception\MetadataException;
+use Tuf\Exception\PotentialAttackException\RollbackAttackException;
```
phenaproxima, line 9: Out of scope, but I'd love to rename this namespace to remove the redundant use of the term "exception". IMHO It should be `Tuf\Exception\PotentialAttack`.

tests/Client/UpdaterTest.php
```diff
@@ -382,4 +389,57 @@ public function providerFileNotFoundExceptions():array
...
+    public function testAttackRepoException(string $fixturesSet, \Exception $expectedException, array $expectedUpdatedVersions): void
+    {
+        // Use the memory storage used so tests can write without permanent
+        // side-effects.
+        $this->localRepo = $this->memoryStorageFromFixture($fixturesSet, 'tufclient/tufrepo/metadata/current');
+        $this->testRepo = new TestRepo($fixturesSet);
+        $this->assertClientRepoVersions(static::getFixtureClientStartVersions($fixturesSet));
+        $updater = $this->getSystemInTest();
+        try {
+            $updater->refresh();
```
phenaproxima, line 416: If I understand this correctly, the rollback attack should be detected, and an exception should be immediately thrown, before any further downloads take place. Right? If that's the case, would it maybe be worth it to find a way to test that the attack is detected *when* we expect it to be, as opposed to "well, eventually a rollback attack is detected, and an exception is thrown", which is what it looks like we are currently doing.

I say this because I know the updater continually is writing things to persistent storage as it refreshes itself. So in theory, if a bad guy could poison the persistent storage _before_ a rollback attack is detected, that could potentially lay the groundwork for further (successful) attacks.

phenaproxima (approved):

tests/Client/UpdaterTest.php
phenaproxima: Nice idea. IMHO we should move `setExceptionOnChange()` into whatever the interface is for the durable storage backends, and rename it to `lock()` (and add an `unlock()` method too). Can be done in a later PR, though.

tedbow: I think if we only need this for testing we should not add it an interface. We should be able to refactor our test with changing our non-test code and affecting any API we may be creating. 

I have added a test for `setExceptionOnChange()`

generate_fixtures.py
davidstrauss: This is a little more ad hoc than I'd prefer long-term; I'd really like to have high-level ways of specifying "attacks." At the same time, I think it's premature to expect to know these patterns yet. I'd like to more forward with this implementation and factor the "attack" fixtures into more descriptive, high-level approaches after we have a few.

davidstrauss (approved):
I haven't reviewed the PHP side, but the Python for the fixtures looks good (if a little low-level for the long-term).

PR #104, opened 2021-02-17
Allow multiple major versions of Symfony and Guzzle

phenaproxima:
Right now, this package cannot be installed alongside Drupal core 8 or 9, since they require versions of Guzzle and Symfony that we don't declare compatibility with. I'd _guess_ that we are not using any APIs which changed between Guzzle 6 and 7 and Symfony 4 and 5 (although I'm by no means the best informed about that), so we should probably allow core-compatible versions of those packages.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,8 +13,8 @@
         "php": ">=7.2.5",
         "ext-json": "*",
         "paragonie/sodium_compat": "^1.13",
-        "symfony/validator": "^5.1",
-        "guzzlehttp/guzzle": "^7.2",
+        "symfony/validator": "^4.4 || ^5.1",
+        "guzzlehttp/guzzle": "^6.5 || ^7.2",
         "myclabs/deep-copy": "^1.10.2"
     },
     "suggest": {
--- a/composer.json
+++ b/composer.json
@@ -13,8 +13,8 @@
         "php": ">=7.2.5",
         "ext-json": "*",
         "paragonie/sodium_compat": "^1.13",
-        "guzzlehttp/guzzle": "^5.3 || ^6.5 || ^7.2",
         "symfony/validator": "^4.4 || ^5.1",
+        "guzzlehttp/guzzle": "^6.5 || ^7.2",
         "myclabs/deep-copy": "^1.10.2"
     },
     "suggest": {
--- a/composer.json
+++ b/composer.json
@@ -13,8 +13,8 @@
         "php": ">=7.2.5",
         "ext-json": "*",
         "paragonie/sodium_compat": "^1.13",
-        "symfony/validator": "^4.4",
-        "guzzlehttp/guzzle": "^6.5",
+        "symfony/validator": "^4.4 || ^5.1",
+        "guzzlehttp/guzzle": "^6.5 || ^7.2",
         "myclabs/deep-copy": "^1.10.2"
     },
     "suggest": {
--- a/composer.json
+++ b/composer.json
@@ -13,8 +13,8 @@
         "php": ">=7.2.5",
         "ext-json": "*",
         "paragonie/sodium_compat": "^1.13",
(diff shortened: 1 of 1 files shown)


xjm:
That's a good point. Let's look up the original issues that added the deps to make sure there wasn't a specific reason?

phenaproxima:
I can definitely speak for Guzzle, because I added it in #81. I did not use Guzzle 7 for any specific reason; that's what happened when I did `composer require guzzlehttp/guzzle`. So as far as I know, we can safely use Guzzle 6.

Symfony, on the other hand, I don't know why, and `git blame` is not helpful in this case. Maybe @tedbow can shed some light on that?

xjm:
https://github.com/php-tuf/php-tuf/pull/81 

We originally were very loose with the Symfony version components we required; I wonder how it got restrictive?

phenaproxima:
> We originally were very loose with the Symfony version components we required; I wonder how it got restrictive?

Is it possible that the same thing happened with Symfony as with Guzzle, and no specific major version was specified?

xjm:
```
        "symfony/validator": "^5.1",
```

So probably the strict requirement was added with the validation PR.

phenaproxima:
Thanks! Skimming #55, I don't see any obvious reason why Symfony 5 was chosen. But I might well be missing something.

mbaynton:
Seems like this could be a problem for potential consumers of this library besides Drupal. Is it too late to talk about solutions like not using dependencies or rewriting their namespaces that would work for anyone?

phenaproxima:
@mbaynton IMHO it would be preferable to just explicitly declare the dependencies and versions we support/allow.

xjm:
We can/should definitely widen the constraints to allow multiple major versions (and test them to make sure they work). AFAIK the only intentional restriction is our PHP version requirement.

phenaproxima:
> We can/should definitely widen the constraints to allow multiple major versions (and test them to make sure they work).

Agreed. However, ideally we'd merge this in this short term to unblock my work on php-tuf/composer-integration, and do a follow-up which adds support for more major versions of our dependencies _and_ workflows to test against those versions.

phenaproxima:
Reverting support for Symfony 3 for now, and opened #105 to discuss that.

phenaproxima:
Same thing with Guzzle 5, it turns out. I've opened #106 to discuss that.

composer.json
```diff
@@ -13,8 +13,8 @@
         "php": ">=7.2.5",
         "ext-json": "*",
         "paragonie/sodium_compat": "^1.13",
-        "symfony/validator": "^5.1",
-        "guzzlehttp/guzzle": "^7.2",
+        "symfony/validator": "^4.4 || ^5.1",
```
xjm: Drupal has an agreement with the Symfony security team that we work with them to provide security coverage for old minor versions of components we use, and we have several Symfony 5.0 components in Drupal 9.0 that we might need to make sure work with this project. Let's see if we can allow Symfony 5.0 for now.

xjm: Symfony 3.4 is actively supported by Symfony until November of this year, but I am fine with a followup for that.

composer.json
```diff
@@ -13,8 +13,8 @@
...
+        "guzzlehttp/guzzle": "^6.5 || ^7.2",
```
xjm, line 17: I believe Guzzle 5 still technically has security coverage even though it's not actively supported. But also OK with a followup to discuss that since it is EOLish. We can decide whether to invest the time or not later.

xjm (changes requested):

composer.json
phenaproxima: Tests are passing with ^5 as the Symfony constraint, so that is probably a good sign. I'm marking this as resolved for now.

composer.json
phenaproxima: Follow-up opened.

xjm (approved):

Issue #107, opened 2021-02-17
Test against all supported major versions of our dependencies


phenaproxima:
In #104, we explicitly declared support for multiple major versions of Symfony and Guzzle. We also need to be sure we're testing against each of those major versions; right now, we're just testing against the latest supported version. Let's adjust our CI configuration to have a matrix of major dependency versions that we can test against.

xjm:
https://github.com/php-tuf/php-tuf/blob/main/.github/workflows/php.yml is the file to change to do this.

PR #108, opened 2021-02-17
Test against all major supported versions of dependencies

phenaproxima:
Addresses #107.


xjm (changes requested):
Nice work! Can we change the names of the jobs though to add the dependency version? See around line 60:
```
  build:
    name: PHP ${{ matrix.php }} test on ${{ matrix.operating-system }}
```

xjm (approved):
Looks good in the test result UI!

PR #110, opened 2021-02-22
GuzzleFileFetcher inadvertently modifies the fetched JSON

phenaproxima:
By calling `JsonNormalizer::asNormalizedJson()` on the data that it fetches, GuzzleFileFetcher inadvertently removes whitespace from the fetched data -- which technically changes the contents of the string, and thus its SHA256 hash. Oops.

After discussion with @tedbow, there doesn't seem to be any obvious reason why the fetcher would need to normalize the JSON; its job is to fetch some number of bytes from the Internet and return them in an unaltered state. So let's fix this.

### DIFF
--- a/tests/Unit/GuzzleFileFetcherTest.php
+++ b/tests/Unit/GuzzleFileFetcherTest.php
@@ -41,10 +41,13 @@ class GuzzleFileFetcherTest extends TestCase
      * @var string
      */
     private $testContent = <<<END
-["oolong",
-  "assam",
-"matcha",
-"herbal"]
+{
+  "tea": ["oolong",
+     "assam",   "matcha",
+"herbal"],
+  "coffee" ["Arabica", "Catuai",
+     "Geisha"
+ ]}
 END;
 
     /**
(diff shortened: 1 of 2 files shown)


tests/Unit/GuzzleFileFetcherTest.php
```diff
@@ -35,9 +35,17 @@ class GuzzleFileFetcherTest extends TestCase
...
+     *
      * @var string
      */
-    private $testContent = '["oolong","assam","matcha","herbal"]';
+    private $testContent = <<<END
+["oolong",
+  "assam",
+"matcha",
+"herbal"]
+END;
```
tedbow, line 48: Could you make the test data an object with keys and the keys be unsorted also? Or have 2 test cases 1 with unsorted keys and 1 with whitespace. 

The previous bug would have showed up in the unsorted keys case also even if there was no whitespace

tedbow (changes requested):

tests/Unit/GuzzleFileFetcherTest.php
phenaproxima: Done.

tedbow (approved):
Looks good!

PR #111, opened 2021-02-22
JsonNormalizer::asNormalizedJson() should not escape slashes

phenaproxima:
JsonNormalizer::asNormalizedJson() currently escapes slashes, which is the default behavior, but this accidentally changes the returned JSON in an undesirable way. Although this could affect anything that goes through asNormalizedJson(), this problem especially affects targets metadata, which contains paths that contain slashes, and ultimately causes signature verification to fail.


tedbow (approved):
Looks good!

PR #114, opened 2021-02-25
Add a method to get known hashes of a target

phenaproxima:
To get #113 done, we'll need a way to get the known hashes for a target. This adds a method for that to TargetsMetadata, plus test coverage.

### DIFF
--- a/tests/Metadata/TargetsMetadataTest.php
+++ b/tests/Metadata/TargetsMetadataTest.php
@@ -27,6 +27,21 @@ class TargetsMetadataTest extends MetaDataBaseTest
         return TargetsMetadata::createFromJson($json);
     }
 
+    /**
+     * @covers ::getHashes
+     *
+     * @return void
+     *   Describe the void.
+     */
+    public function testGetHashes(): void
+    {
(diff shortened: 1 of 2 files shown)


tests/Metadata/TargetsMetadataTest.php
```diff
@@ -48,6 +67,8 @@ public function providerExpectedField() : array
         $data[] = ['signed:delegations:roles:0:paths'];
         $data[] = ['signed:delegations:roles:0:terminating'];
         $data[] = ['signed:delegations:roles:0:threshold'];
+        $target = $this->getFixtureNestedArrayFirstKey($this->validJson, ['signed', 'targets']);
+        $data[] = ["signed:targets:$target:hashes"];
```
tedbow, line 71: could this to providerValidField also

tedbow (changes requested):
looks good. just need to add `providerValidField` changes

tests/Metadata/TargetsMetadataTest.php
phenaproxima: Done!

tedbow (approved):
Great!

PR #117, opened 2021-03-01
Split RepoFileFetcherInterface::fetchFile() into two separate methods

phenaproxima:
It's become clear that we will probably need different fetching procedures when downloading metadata vs. downloading target files. For that reason, it makes sense to have two different methods of RepoFileFetcher for these two operations. Initially they will both do the same thing (as they do now), but this API change will allow them to diverge in the future.


tedbow (approved):
Looks good!

PR #118, opened 2021-03-03
Use progress callback and stream size in GuzzleFileFetcher to enforce maxBytes

phenaproxima:
This is just an internal refactoring -- when using cURL as its backend, Guzzle supports a progress callback function which can abort the transfer if too many bytes come in. Let's use that to enforce the maximum download size.

Additionally, Guzzle's stream system lets us directly get the size of the stream without necessarily having to read it into memory. This would be quite beneficial for, say, larger file downloads. So I propose we leverage that when checking the transfer size after the fact (which we still need to do, in case cURL is not used as the backend). As a fallback, we should keep the existing code which reads N bytes into memory and verifies the maximum download size.

### DIFF
--- a/src/Client/GuzzleFileFetcher.php
+++ b/src/Client/GuzzleFileFetcher.php
@@ -6,6 +6,7 @@ use GuzzleHttp\Client;
 use GuzzleHttp\ClientInterface;
 use GuzzleHttp\Exception\ClientException;
 use GuzzleHttp\Promise\PromiseInterface;
+use GuzzleHttp\RequestOptions;
 use Psr\Http\Message\ResponseInterface;
 use Tuf\Exception\DownloadSizeException;
 use Tuf\Exception\RepoFileNotFound;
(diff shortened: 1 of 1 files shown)


src/Client/GuzzleFileFetcher.php
```diff
@@ -96,14 +108,23 @@ private function onFulfilled(string $fileName, int $maxBytes): \Closure
...
         return function (ResponseInterface $response) use ($fileName, $maxBytes) {
             $body = $response->getBody();
-            $contents = $body->read($maxBytes);
+            $size = $body->getSize();
 
-            // If we reached the end of the stream, we didn't exceed the maximum
-            // number of bytes.
-            if ($body->eof() === true) {
-                return $contents;
+            if (isset($size)) {
```
tedbow, line 113: I am unclear when this would not be set?

phenaproxima: `StreamInterface` makes an allowance for streams whose size is not known:
```php
    /**
     * Get the size of the stream if known.
     *
     * @return int|null Returns the size in bytes if known, or null if unknown.
     */
    public function getSize();
```
So in this case, I'm just respecting the documented behavior of the API.

tedbow (approved):
Looks good!

PR #119, opened 2021-03-03
Promises returned by RepoFileFetcherInterface should wrap streams

phenaproxima:
As I've been discovering, we need to be able to handle two kinds of data retrieval from the Internet: strings, and files. I've been trying to figure out how to do this consistently -- in other words, we already use promises for all our HTTP transactions, but what should the promises wrap around if some of the transactions should return strings, and others should just download files without returning their contents?

Having pondered it for a couple of days, I believe the answer is streams, which Guzzle already supports using its own PSR-7 implementation. RepoFileFetcherInterface's promises should always wrap around an instance of StreamInterface, which could be referring either to a string in memory, or a file downloaded on disk (whose path we can retrieve as part of the stream metadata). This adds a further layer of normalization and consistency to our API, and leverages another library (in addition to Guzzle's promises library) that we have as an indirect dependency.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,8 @@
         "paragonie/sodium_compat": "^1.13",
         "symfony/validator": "^4.4 || ^5",
         "guzzlehttp/guzzle": "^6.5 || ^7.2",
-        "myclabs/deep-copy": "^1.10.2"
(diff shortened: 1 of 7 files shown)


tedbow (approved):
Looks great!

PR #120, opened 2021-03-03
Use hash_file() to verify file streams where appropriate

phenaproxima:
Now that we have adopted streams (in #119), we need to support the ability to call `hash_file()` for verification of downloaded data, so that we don't _always_ need to read the entire contents of the stream into memory to compute the hash.

According to https://www.php.net/stream_get_meta_data, streams can have a `uri` metadata property which refers to their source -- which could be a file path, if that's where Guzzle saved the downloaded data. We should leverage that property to see if the stream is coming from a (downloaded) file, and use hash_file() in that case.

### DIFF
--- a/src/Client/Updater.php
+++ b/src/Client/Updater.php
@@ -546,9 +546,20 @@ class Updater
         $length = $targetsMetaData->getLength($target) ?? static::MAXIMUM_DOWNLOAD_BYTES;
 
         $verify = function (StreamInterface $stream) use ($target, $hashes) {
-            $content = $stream->getContents();
+            // If the stream has a URI that refers to a file, use hash_file() to
+            // verify it. Otherwise, read the entire stream as a string and use
+            // hash() to verify it.
+            $uri = $stream->getMetadata('uri');
+            if ($uri && file_exists($uri)) {
+                $content = $uri;
+                $verifier = 'hash_file';
+            } else {
+                $content = $stream->getContents();
+                $verifier = 'hash';
+            }
+
             foreach ($hashes as $algo => $hash) {
-                if ($hash !== hash($algo, $content)) {
+                if ($hash !== $verifier($algo, $content)) {
                     throw new InvalidHashException("Invalid $algo hash for $target");
                 }
             }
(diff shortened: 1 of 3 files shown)


src/Client/Updater.php
```diff
@@ -546,9 +546,20 @@ public function download(string $target): PromiseInterface
...
+                $content = $uri;
+                $verifier = 'hash_file';
+            } else {
+                $content = $stream->getContents();
+                $verifier = 'hash';
+            }
+
             foreach ($hashes as $algo => $hash) {
-                if ($hash !== hash($algo, $content)) {
+                if ($hash !== $verifier($algo, $content)) {
```
tedbow, line 562: This seems weird to use a variable as the function call if we don't have to.

Also is `$content` really content if it was set to `$uri`. Why not something like this

           $uri = $stream->getMetadata('uri');
            if (!$uri || !file_exists($uri)) {
                $content = $stream->getContents();
            }

            foreach ($hashes as $algo => $hash) {
                $streamHash = isset($content) ? hash($algo, $content) : hash_file($algo, $uri);
                if ($hash !== $streamHash) {
                    throw new InvalidHashException("Invalid $algo hash for $target");
                }
            }

tests/Client/UpdaterTest.php
```diff
@@ -114,10 +115,20 @@ public function testVerifiedDownload(): void
         $this->testRepo = new TestRepo($fixturesSet);
         $updater = $this->getSystemInTest();
 
-        $testFileContents = file_get_contents(__DIR__ . '/../../fixtures/TUFTestFixtureSimple/tufrepo/targets/testtarget.txt');
+        $testFilePath = realpath(__DIR__ . '/../../fixtures/TUFTestFixtureSimple/tufrepo/targets/testtarget.txt');
```
tedbow, line 118: I didn't catch this before but we should use `\Tuf\Tests\TestHelpers\UtilsTrait::getFixturesRealPath` here.

tedbow (changes requested):
Couple small suggestions but looks good

src/Client/Updater.php
phenaproxima: Fixed this by moving the existing logic into the loop and removing the function name variable.

tedbow (approved):
Looks good!

PR #121, opened 2021-03-04
Add a StreamInterface implementation that is aware of the HTTP response

phenaproxima:
In some cases, consumers of TUF (right now, that means php-tuf/composer-integration) will need information about the HTTP response that came back from a request. Right now, though, our promises wrap streams -- which is good. The streams have zero awareness about the response (or even that they were produced by an HTTP transaction) -- which is also good. But for integrating with Composer (and possibly other things), it's too limiting.

I propose to add a simple adapter class which implements `StreamInterface`, and wraps around both a stream _and_ the response which produced the stream. That way, calling code can examine the response if it needs to, but also continue to use the resulting stream as any other.

### DIFF
--- a/src/Client/GuzzleFileFetcher.php
+++ b/src/Client/GuzzleFileFetcher.php
@@ -124,7 +124,7 @@ class GuzzleFileFetcher implements RepoFileFetcherInterface
                 }
                 $body->rewind();
             }
-            return new ResponseAwareStream($body, $response);
+            return new ResponseStream($response);
         };
     }
 
--- a/src/Client/GuzzleFileFetcher.php
+++ b/src/Client/GuzzleFileFetcher.php
@@ -124,7 +124,7 @@ class GuzzleFileFetcher implements RepoFileFetcherInterface
                 }
                 $body->rewind();
             }
-            return $body;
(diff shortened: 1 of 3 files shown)


tedbow:
Can you explain why composer integration needs the response?

phenaproxima:
Yes. What we basically need to do is substitute the Composer's `HttpDownloader` in favor of TUF's download mechanisms, which include transparent length and hash verification. This is pretty tricky because Composer's `HttpDownloader` is not very abstracted at all, and Composer seems to be pretty closely coupled to it. That means is that TUF, and especially our plugin, needs to be able to emulate the behavior of `HttpDownloader` as closely as possible.

One of the main outputs of `HttpDownloader` is `Composer\Util\Http\Response` objects. These are what you think they are, just very lightweight. In my opinion, the more "complete" we can make these, the better off we will be; Composer uses things like response headers and such to determine some of its behaviors after it downloads a file. So we should give Composer as many details as possible about the response. But right now TUF makes that impossible, so it needs an agnostic way to provide response information to consumers (like our plugin) that might need it.

tedbow (approved):
Looks good!

PR #122, opened 2021-03-04
GuzzleFileFetcher::create() should allow HTTP URIs

phenaproxima:
Right now, GuzzleFileFetcher::create() throws an exception if you give it a non-HTTPS URI. Since HTTPS is not the main form of security in TUF, this is needlessly restrictive.


tedbow (approved):
👍

PR #123, opened 2021-03-04
Allow file fetchers to accept idiosyncratic options

phenaproxima:
As with #121, this is a minor API change to support Composer integration in a consumer-agnostic way. This will allow the plugin to pass the `sink` option to the Guzzle file fetcher, which will make Guzzle download a target to a specific path (Composer's `HttpDownloader` has a similar functionality, so this lets us parallel that).

### DIFF
--- a/src/Client/GuzzleFileFetcher.php
+++ b/src/Client/GuzzleFileFetcher.php
@@ -65,15 +65,15 @@ class GuzzleFileFetcher implements RepoFileFetcherInterface
     /**
      * {@inheritDoc}
      */
-    public function fetchTarget(string $fileName, int $maxBytes): PromiseInterface
+    public function fetchTarget(string $fileName, int $maxBytes, array $options = []): PromiseInterface
     {
-        return $this->fetchFile($fileName, $maxBytes);
+        return $this->fetchFile($fileName, $maxBytes, $options);
     }
 
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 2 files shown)


src/Client/RepoFileFetcherInterface.php
```diff
@@ -16,14 +16,16 @@ interface RepoFileFetcherInterface
...
+     *   (optional) Additional fetcher-specific options.
      *
      * @return \GuzzleHttp\Promise\PromiseInterface
      *   A promise representing the eventual result of the operation. If
      *   successful, the promise should wrap around an instance of
      *   \Psr\Http\Message\StreamInterface, which provides a stream of the
      *   retrieved data.
      */
-    public function fetchMetaData(string $fileName, int $maxBytes): PromiseInterface;
+    public function fetchMetaData(string $fileName, int $maxBytes, array $options = []): PromiseInterface;
```
tedbow, line 28: @phenaproxima what do you think about not adding `$options` to the interface methods but just GuzzleFileFetcher since this is an optional argument it would work. because other fetchers might want to pass there args in a different manner. Like if a different backend had an options class. or even if they just wanted to add a couple of boolean flag arguments

phenaproxima: Sure, that works! Nice idea.

tedbow (approved):
👀 📖 👍

PR #124, opened 2021-03-05
Updater::download() should return a rejected promise if the target is invalid

phenaproxima:
`Updater::download()` is not really doing the right thing if the requested target is either non-existent or has no trusted hashes. It's _supposed_ to return a promise, but instead it's immediately throwing an exception, which means calling code needs to have a try-catch around a promise. That's nuts. Instead, it should just immediately return a rejected promise, which makes more sense given how promises are supposed to work.


tedbow (approved):
Approved!

PR #125, opened 2021-03-08
Support path prefixes for metadata and targets in GuzzleFileFetcher

phenaproxima:
The Python repository tool generates the repository in a specific way: TUF metadata goes into a `/metadata` directory, and the actual targets go into `/targets`. GuzzleFileFetcher should make it easier to deal with this layout by optionally remembering the path prefixes for these things.

### DIFF
--- a/src/Client/GuzzleFileFetcher.php
+++ b/src/Client/GuzzleFileFetcher.php
@@ -24,14 +24,34 @@ class GuzzleFileFetcher implements RepoFileFetcherInterface
     private $client;
 
     /**
+     * The path prefix for metadata.
+     *
+     * @var string|null
+     */
+    private $metaDataPrefix;
+
+    /**
+     * The path prefix for targets.
+     *
+     * @var string|null
+     */
+    private $targetsPrefix;
+
+    /**
      * GuzzleFileFetcher constructor.
      *
      * @param \GuzzleHttp\ClientInterface $client
      *   The HTTP client.
+     * @param string|null $metaDataPrefix
+     *   (optional) The path prefix for metadata.
+     * @param string|null $targetsPrefix
+     *   (optional) The path prefix for targets.
      */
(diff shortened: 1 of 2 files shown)


src/Client/GuzzleFileFetcher.php
```diff
@@ -23,15 +23,35 @@ class GuzzleFileFetcher implements RepoFileFetcherInterface
...
      *
      * @param \GuzzleHttp\ClientInterface $client
      *   The HTTP client.
+     * @param string|null $metaDataPrefix
+     *   (optional) The path prefix for metadata.
+     * @param string|null $targetsPrefix
+     *   (optional) The path prefix for targets.
      */
-    public function __construct(ClientInterface $client)
+    public function __construct(ClientInterface $client, string $metaDataPrefix = null, string $targetsPrefix = null)
```
tedbow, line 50: why not default these to `metadata` and `targets`?

phenaproxima: I'd say, because that would be an opinion, and this fetcher should be very agnostic as to what the layout actually is. Defaulting them means that the fetcher is assuming you used our Python code to generate your repository. We'd be putting future consumers of this library in a position to _undo_ our defaults rather than take advantage of these helpers.

tedbow: That means we are favoring not having an opinion over the consumer not having to know the internals of how the TUF clients stores  stuff/works.

I would say it is more important for the caller to have to know as little as possible with the ability of non-standard users to be able to override these paths if needed.

phenaproxima: > That means we are favoring not having an opinion over the consumer not having to know the internals of how the TUF clients stores stuff/works.

If we default these, the caller still needs to know how the repository is laid out on the server side.

tedbow (approved):
Looks good now!

PR #127, opened 2021-03-19
Allow GuzzleFileFetcher::fetchTarget() to bypass prefixing

phenaproxima:
Our plan to sign Packagist packages means that we need to be able to sign targets that live on a separate server. In order to make that work, we need to be able to pass full URLs to GuzzleFileFetcher::fetchTarget(), bypassing the prefixing that we do.


tedbow (approved):
Looks good. thanks for making the follow-up issue

Issue #128, opened 2021-03-19
Figure out how to sign remote packages


phenaproxima:
#127 introduces a small under-the-hood change to make it possible for php-tuf/composer-integration to use TUF to download remote targets. A "remote target" is any target which is NOT hosted on the same server as the TUF metadata. Essentially it's a concept of treating an arbitrary URL as a signed, validated target.

As far as I can tell, we're going to _need_ to support something like this in order for us to sign packages which come from Packagist, rather than drupal.org. But it doesn't look like the Python repository tool (which we use to create TUF metadata from arbitrary, local target files) has a way to sign a remote URL. It seems to expect everything it signs to be a local file. Which makes sense, but doesn't help us.

The workaround I'm implementing to sign remote targets is:

1. Download the remote file from some URL that you trust.
2. Rename the downloaded file after the SHA256 hash of the URL from which you retrieved it. So basically, `somefile.zip` --> `sha256HashOfTheUrl`.
3. Modify generate.py to sign that target.

Then, the plugin is doing this:

1. Use a custom package loader to store the SHA256 hash of the package's dist URL -- i.e., determine the name of the target.
2. Use a custom file fetcher, extending the Guzzle fetcher, which stores a map of target names to "real" URLs.
3. At download time, ensure that the real URL is downloaded and run through TUF's validation.

In other words: TUF knows about the remote target by a SHA256 hash. But internally, the fetcher knows to retrieve a different URL. Ugly, I know...but until the Python repository tool can sign remote URLs natively, this is necessary. I would massively prefer if TUF could just treat URLs like regular targets!

davidstrauss:
> As far as I can tell, we're going to need to support something like this in order for us to sign packages which come from Packagist, rather than drupal.org.

I was also considering something like configuring Drupal.org Fastly to proxy.

phenaproxima:
I think this is pretty much resolved at this point.

This library (and, as best I can figure, TUF itself) doesn't care where the data actually originates. As long as it matches the verified hashes available in the TUF metadata, it's trustworthy.

Signing remote packages is also not something this library will need to worry about. This merely verifies signed data, but doesn't sign anything itself. The Composer plugin uses the scheme of identifying a signed package by the target name with the scheme `VENDOR/NAME/NORMALIZED_VERSION`.

I'm closing this out.

PR #129, opened 2021-03-23
Allow targets to be downloaded from arbitrary URLs

phenaproxima:
This is related to #128. It may be necessary for us to sign a package using a local copy of the file, but actually _download_ that file, in a runtime situation, from some arbitrary URL. This PR makes the GuzzleFileFetcher (and only that fetcher) support mapping a local target path to an arbitrary URL.

### DIFF
--- a/src/Client/GuzzleFileFetcher.php
+++ b/src/Client/GuzzleFileFetcher.php
@@ -37,6 +37,11 @@ class GuzzleFileFetcher implements RepoFileFetcherInterface
      */
     private $targetsPrefix;
 
+    /**
+     * A map of targets to the URL from which they should be downloaded.
+     *
+     * @var array
+     */
     private $targetUrls = [];
 
     /**
@@ -75,6 +80,17 @@ class GuzzleFileFetcher implements RepoFileFetcherInterface
         return new static($client, $metaDataPrefix, $targetsPrefix);
     }
 
+    /**
+     * Maps a target to an arbitrary URL.
+     *
+     * @param string $target
+     *   The target to map.
+     * @param string|null $url
+     *   The URL from which the target should be downloaded, or null to unmap
+     *   a previously set URL.
+     *
+     * @return $this
+     */
     public function setTargetUrl(string $target, ?string $url): self
     {
         if (isset($url)) {
--- a/src/Client/GuzzleFileFetcher.php
+++ b/src/Client/GuzzleFileFetcher.php
@@ -75,9 +75,13 @@ class GuzzleFileFetcher implements RepoFileFetcherInterface
         return new static($client, $metaDataPrefix, $targetsPrefix);
     }
 
-    public function setTargetUrl(string $target, string $url): self
+    public function setTargetUrl(string $target, ?string $url): self
     {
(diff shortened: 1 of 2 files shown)


src/Client/GuzzleFileFetcher.php
```diff
@@ -84,8 +84,10 @@ public function fetchMetaData(string $fileName, int $maxBytes): PromiseInterface
     /**
      * {@inheritDoc}
      */
-    public function fetchTarget(string $fileName, int $maxBytes, array $options = []): PromiseInterface
+    public function fetchTarget(string $fileName, int $maxBytes, array $options = [], string $url = null): PromiseInterface
```
tedbow, line 87: Even though we are using `@inheritdoc` can we still document the extra `@param`'s somehow?

src/Client/GuzzleFileFetcher.php
```diff
@@ -84,8 +84,10 @@ public function fetchMetaData(string $fileName, int $maxBytes): PromiseInterface
...
     {
+        // Allow calling code to download the target from an arbitrary URL.
+        $fileName = $url ?? $fileName;
```
tedbow, line 90: Should we assign to a new more generic var like `$location` or something?

tedbow (changes requested):
Looks good. Just a couple suggestions

src/Client/GuzzleFileFetcher.php
phenaproxima: Sure! Documentation added.

src/Client/GuzzleFileFetcher.php
phenaproxima: Nice idea. Renamed to `$location`.

src/Client/GuzzleFileFetcher.php
```diff
@@ -83,39 +83,59 @@ public function fetchMetaData(string $fileName, int $maxBytes): PromiseInterface
...
+        // Allow calling code to download the target from an arbitrary URL.
+        $location = $url ?? $fileName;
+        // If $location isn't a full URL, treat it as a relative path and prefix
         // it with $this->targetsPrefix.
         // @todo Revisit the need for this bypass once
         // https://github.com/php-tuf/php-tuf/issues/128 is resolved. This was
         // originally added because of the workaround described there.
-        if (parse_url($fileName, PHP_URL_HOST) === null) {
-            $fileName = $this->targetsPrefix . $fileName;
+        if (parse_url($location, PHP_URL_HOST) === null) {
```
tedbow, line 103: why don't we do this only if $url is provided and through an error if 
`parse_url($url, PHP_URL_HOST) === null`
? 
and then only add the prefix if $url is not provided.

Right now you could provide a URL string to $fileName and this would still work.

phenaproxima: Discussed and I think your insight makes sense. We don't really need the ability to bypass the prefixing, because this PR introduces a way to pass a full URL. So I removed all of this code and simplified it.

tedbow (approved):
Looks good!

PR #131, opened 2021-04-01
Make InvalidHashException aware of the offending stream

phenaproxima:
In order to get around the fact that the Composer integration plugin doesn't handle 304 responses, we need to be able to convert an InvalidHashException caused by a 304 to a legitimate response object. In order to do *that*, we need some way for calling code to access the stream that caused the InvalidHashException. This PR adds such an accessor.

### DIFF
--- a/src/Exception/PotentialAttackException/InvalidHashException.php
+++ b/src/Exception/PotentialAttackException/InvalidHashException.php
@@ -9,18 +9,18 @@ use Tuf\Exception\TufException;
  * Indicates an invalid hash was computed for a downloaded target.
  *
  * This exception provides access to the underlying stream which caused the
- * error. The stream's contents are NOT trusted by TUF, so any code which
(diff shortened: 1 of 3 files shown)


tedbow (approved):
Looks good!

src/Exception/PotentialAttackException/InvalidHashException.php
```diff
@@ -2,11 +2,62 @@
...
+ * error. The stream's contents are NOT trusted by TUF, so any code which
+ * interacts with this exception, or the stream, should do so with great caution
+ * and treat the stream contents as unsafe.
  */
 class InvalidHashException extends TufException
 {
+    /**
+     * An untrusted stream object pointing to the downloaded target.
+     *
+     * WARNING: The contents of the stream are NOT trusted by TUF! Any code
```
tedbow, line 21: It is not only that it is not trusted. It actually "failed TUF validation". maybe we should say that.

PR #132, opened 2021-04-02
Stop using deep_copy() function

phenaproxima:
For whatever reason, in some circumstances Composer seems to have a problem resolving the deep_copy() function. It looks like it's a pretty thin wrapper anyway; I propose we use the class reference directly.


tedbow (approved):
looks good!

PR #133, opened 2021-04-05
TargetsMetadata should throw a TUF-specific exception for undefined targets

phenaproxima:
Right now, if the updater requests metadata about a specific target and that target is unknown, we throw InvalidArgumentException. Although that technically works, it's very ambiguous to the calling code; it has no real way to know that the exception was thrown by TUF, for TUF-specific reasons. We should use our already-existing NotFoundException instead, which will be a lot clearer to consumers of TUF.

### DIFF
--- a/tests/Metadata/TargetsMetadataTest.php
+++ b/tests/Metadata/TargetsMetadataTest.php
@@ -2,6 +2,7 @@
 
 namespace Tuf\Tests\Metadata;
 
+use Tuf\Exception\NotFoundException;
 use Tuf\Metadata\MetadataBase;
 use Tuf\Metadata\TargetsMetadata;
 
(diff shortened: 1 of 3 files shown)


tests/Metadata/TargetsMetadataTest.php
```diff
@@ -48,15 +49,15 @@ public function testGetHashesAndLength(): void
...
+            $this->assertSame("Target not found: void.txt", $e->getMessage());
         }
 
         try {
             $metadata->getLength('void.txt');
             $this->fail('Exception was not thrown for an invalid target.');
-        } catch (\InvalidArgumentException $e) {
-            $this->assertSame("Unknown target: 'void.txt'", $e->getMessage());
+        } catch (NotFoundException $e) {
+            $this->assertSame("Target not found: void.txt", $e->getMessage());
```
tedbow, line 60: Small nit is removing the single quote around the file name intentional?

TravisCarden (approved):

tests/Metadata/TargetsMetadataTest.php
phenaproxima: It is not. NotFoundException has its own logic to determine how the exception message is formatted, and it doesn't put the single quotes in.

PR #135, opened 2021-04-06
Updater::refresh() should return early if already refreshed

phenaproxima:
Rather than requiring calling code to check `$this->isRefreshed`, `Updater::refresh()` should be smart enough to return early if it's already been refreshed. To wit, it should also support a `$force` flag to make it refresh from the server anyway.

### DIFF
--- a/src/Client/Updater.php
+++ b/src/Client/Updater.php
@@ -132,13 +132,24 @@ class Updater
      * @todo The Python implementation has an optional flag to "unsafely update
      *     root if necessary". Do we need it?
      *
+     * @param bool $force
+     *   (optional) If false, return early if this updater has already been
+     *   refreshed. Defaults to false.
+     *
      * @return boolean
      *     TRUE if the data was successfully refreshed.
      *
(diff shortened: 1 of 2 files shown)


src/Client/Updater.php
```diff
@@ -535,9 +546,8 @@ private function fetchFile(string $fileName, int $maxBytes = self::MAXIMUM_DOWNL
      */
     public function download(string $target, ...$extra): PromiseInterface
     {
-        if (!$this->isRefreshed) {
-            $this->refresh();
-        }
+        $this->refresh();
```
tedbow, line 549: I don't think we need this change be change because `refresh()` method will take care of returning early since we are not passing `$force`

tedbow (changes requested):
Looks good. Just 1 suggested change

src/Client/Updater.php
tedbow: NVB I had the  +/- reversed in my head. 🤦🏼‍♂️

tedbow (approved):
Looks great!

PR #137, opened 2021-04-06
Adds getRole method 

tedbow:
Role is not always the same as type. Delegate roles will be the type targets' but role specified in the delegation.

### DIFF
--- a/src/Metadata/TargetsMetadata.php
+++ b/src/Metadata/TargetsMetadata.php
@@ -32,7 +32,7 @@ class TargetsMetadata extends MetadataBase
      * @param string|null $roleName
      *   The role name if not the same as the type.
      */
-    public static function createFromJson(string $json, ?string $roleName = null)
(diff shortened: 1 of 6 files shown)


src/Metadata/TargetsMetadata.php
```diff
@@ -19,6 +19,26 @@ class TargetsMetadata extends MetadataBase
...
+     */
+    private $role;
+
+    /**
+     * {@inheritdoc}
+     *
+     * @param string|null $roleName
+     *   The role name if not the same as the type.
+     */
+    public static function createFromJson(string $json, ?string $roleName = null)
```
phenaproxima, line 35:
```suggestion
    public static function createFromJson(string $json, string $roleName = null)
```

phenaproxima (commented):
Looks okay to me, just one minor suggestion. I also reviewed the delegation stuff in the TUF spec, and I think this seems clear.

phenaproxima (approved):

PR #139, opened 2021-04-06
Updater should validate the length of a downloaded stream

phenaproxima:
Right now, it's up to the file fetchers to validate the length of the data they retrieve. Since bounds checking is important in TUF, this is something that the Updater should do itself, by acting on the streams returned from the file fetchers.

### DIFF
--- a/src/Client/Updater.php
+++ b/src/Client/Updater.php
@@ -560,6 +560,8 @@ class Updater
                 throw $error;
             }
         } else {
+            // @todo Handle non-seekable streams.
+            $data->rewind();
             $data->read($maxBytes);
 
             // If we reached the end of the stream, we didn't exceed the
@@ -567,7 +569,6 @@ class Updater
             if ($data->eof() === false) {
                 throw $error;
             }
-            // @todo Handle non-seekable streams with unknown size.
             $data->rewind();
         }
     }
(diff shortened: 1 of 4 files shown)


src/Client/Updater.php
```diff
@@ -528,7 +529,47 @@ private static function hasRotatedKeys(RootMetadata $previousRootData, RootMetad
...
+            }
+        } else {
+            $data->read($maxBytes);
+
+            // If we reached the end of the stream, we didn't exceed the
+            // maximum number of bytes.
+            if ($data->eof() === false) {
+                throw $error;
+            }
+            // @todo Handle non-seekable streams with unknown size.
```
tedbow, line 570: Is there a way we can tell if a stream is non-seekable here? We could just throw an exception here and say this type of stream is unsupported for now.

phenaproxima: The docs of `StreamInterface::rewind()` specify that non-seekable streams will throw an exception when that method is called, so we don't need to do anything additional here. But in the future, it would be nice to handle non-seekable streams gracefully if we can.

phenaproxima: Also, `StreamInterface` has an `isSeekable()` method, precisely for this kind of introspection.

tedbow: Should we call `rewind()` before `read($maxBytes)` to ensure we are at the beginning before test to see if it gets us to the end of the stream. We would still have to call `rewind()` after also.

phenaproxima: Yeah, that makes sense. I'll add it.

tedbow (approved):
Looks good!

PR #140, opened 2021-04-06
Add helper methods to get delegated info

tedbow:
Adds TargetsMetadata::getDelegatedKeys() and TargetsMetadata::getDelegatedRoles()


phenaproxima (approved):
Looks okay to me. These are just simple accessors for now, and that's fine; in the future it may make sense to have these return new instances of TargetsMetadata (or some subclass thereof which knows about its delegation).

PR #141, opened 2021-04-06
Search for targets in delegated target files

tedbow:
Implements 5.4.5 of the TUF spec. This is the final portion of the TUF client spec!

Created follow up issue:  https://github.com/php-tuf/php-tuf/issues/142 to test nested delegation


src/Client/Updater.php
```diff
@@ -634,4 +638,51 @@ public function download(string $target, ...$extra): PromiseInterface
...
     }
+
+    /**
+     * @param string $target
+     *   The path of the target file. Needs to be known to the most recent
+     *   targets metadata downloaded in ::refresh().
+     */
+    private function getMetadataForTarget(string $target, SnapshotMetadata $snapshotMetadata, ?TargetsMetadata $targetsMetadata = null): ?TargetsMetadata
+    {
+        static $searchedRoles = [];
```
phenaproxima, line 649: I know static properties are the same for all instances of a class, but how does it work for static variables in class methods? I just want to be certain this won't get polluted by other instances of `Updater` running in the same memory space (i.e., what the Composer plugin does).

src/Client/Updater.php
```diff
@@ -590,10 +590,14 @@ public function download(string $target, ...$extra): PromiseInterface
     {
         $this->refresh();
 
-        // @todo Handle the possibility that the target's metadata might not be
-        // in targets.json.
-        // @see https://github.com/php-tuf/php-tuf/issues/116
-        $targetsMetaData = TargetsMetadata::createFromJson($this->durableStorage['targets.json']);
+        $snapShotMetadata = SnapshotMetadata::createFromJson($this->durableStorage['snapshot.json']);
```
phenaproxima, line 593: Nit: Capitalization inconsistency between "targetsMetaData" and "snapShotMetadata". IMHO this should simply be "targetsMetadata" and "snapshotMetadata".

src/Client/Updater.php
```diff
@@ -634,4 +638,51 @@ public function download(string $target, ...$extra): PromiseInterface
         return $this->repoFileFetcher->fetchTarget($target, $length, ...$extra)
             ->then($verify);
     }
+
+    /**
+     * @param string $target
+     *   The path of the target file. Needs to be known to the most recent
+     *   targets metadata downloaded in ::refresh().
+     */
+    private function getMetadataForTarget(string $target, SnapshotMetadata $snapshotMetadata, ?TargetsMetadata $targetsMetadata = null): ?TargetsMetadata
```
phenaproxima, line 647: The other two parameters need to be documented here as well (plus a description).

src/Client/Updater.php
```diff
@@ -634,4 +638,51 @@ public function download(string $target, ...$extra): PromiseInterface
...
+        static $searchedRoles = [];
+        if ($targetsMetadata === null) {
+            $targetsMetadata = TargetsMetadata::createFromJson($this->durableStorage['targets.json']);
+            $searchedRoles = [];
+        }
+        if ($targetsMetadata->hasTarget($target)) {
+            return $targetsMetadata;
+        } else {
+            $delegatedKeys = $targetsMetadata->getDelegatedKeys();
+            foreach ($delegatedKeys as $delegatedKey) {
```
phenaproxima, line 658: Is this something that should maybe happen in `refresh()`?

src/Client/Updater.php
```diff
@@ -634,4 +638,51 @@ public function download(string $target, ...$extra): PromiseInterface
...
+                $this->keyDB->addKey($delegatedKey);
+            }
+            $delegatedRoles = $targetsMetadata->getDelegatedRoles();
+            foreach ($delegatedRoles as $delegatedRole) {
+                $delegatedRoleName = $delegatedRole['name'];
+                if (in_array($delegatedRole, $searchedRoles)) {
```
phenaproxima, line 664:
```suggestion
                if (in_array($delegatedRole, $searchedRoles, true)) {
```

src/Client/Updater.php
```diff
@@ -634,4 +638,51 @@ public function download(string $target, ...$extra): PromiseInterface
...
+                $newTargetsContent = $this->fetchFile($targetsFileName);
+                $newTargetsData = TargetsMetadata::createFromJson($newTargetsContent, $delegatedRoleName);
+                $snapshotMetadata->verifyNewMetaData($newTargetsData);
+                $this->checkSignatures($newTargetsData);
+                // TUF-SPEC-v1.0.9 Section 5.4.3
+                static::checkFreezeAttack($newTargetsData, $this->getCurrentTime());
+                $newTargetsData->setIsTrusted(true);
+                // TUF-SPEC-v1.0.9 Section 5.4.4
+                $this->durableStorage[$targetsFileName] = $newTargetsContent;
+                $newTargetsData->hasTarget($target);
```
phenaproxima, line 682: `hasTarget()` returns a boolean, and we're not using the return value here. This looks like it might be incomplete...?

phenaproxima (changes requested):

src/Client/Updater.php
tedbow: Yes, this would have been a problem. No longer using static vars here. Good catch.

src/Client/Updater.php
tedbow: fixed

src/Client/Updater.php
tedbow: fix was fixed in another PR

src/Client/Updater.php
tedbow: fixed

src/Client/Updater.php
tedbow: yep fixed

src/Client/Updater.php
```diff
@@ -639,12 +636,76 @@ public function download(string $target, ...$extra): PromiseInterface
...
+     *   The targets metadata to search or null. If null then the search will
+     *   start at the top level 'targets.json' file.
+     * @param string[] $searchedRoles
+     *   The roles that have already been searched.
+     *
+     * @return \Tuf\Metadata\TargetsMetadata|null
+     *   The target metadata with a match for the target, or null no match is
+     *   found.
+     */
+    private function getMetadataForTarget(string $target, ?TargetsMetadata $targetsMetadata = null, array $searchedRoles = []): ?TargetsMetadata
```
phenaproxima, line 661: I think I might need this to be protected, since it will be necessary for the Composer plugin to look up information about specific targets.

src/Client/Updater.php
```diff
@@ -578,11 +578,7 @@ protected function verify(string $target, StreamInterface $data): void
     {
         $this->refresh();
 
-        // @todo Handle the possibility that the target's metadata might not be
-        // in targets.json.
-        // @see https://github.com/php-tuf/php-tuf/issues/116
-        $targetsMetadata = TargetsMetadata::createFromJson($this->durableStorage['targets.json']);
-
+        $targetsMetadata = $this->getMetadataForTarget($target);
```
phenaproxima, line 581: What happens if $targetsMetadata comes back NULL? Should we throw an exception?

src/Client/Updater.php
```diff
@@ -644,6 +640,70 @@ public function download(string $target, ...$extra): PromiseInterface
...
+     *   The targets metadata to search or null. If null then the search will
+     *   start at the top level 'targets.json' file.
+     * @param string[] $searchedRoles
+     *   The roles that have already been searched.
+     *
+     * @return \Tuf\Metadata\TargetsMetadata|null
+     *   The target metadata with a match for the target, or null no match is
+     *   found.
+     */
+    protected function getMetadataForTarget(string $target, ?TargetsMetadata $targetsMetadata = null, array $searchedRoles = []): ?TargetsMetadata
```
phenaproxima, line 659: Wondering if we should document $searchedRoles as being internal use only, since AFAIK it's used for recursion control. I can't think of any reason why it should be passed by the calling code.

src/Client/Updater.php
```diff
@@ -644,6 +640,70 @@ public function download(string $target, ...$extra): PromiseInterface
...
+    {
+        if ($targetsMetadata === null) {
+            // If no target metadata is provided then start searching with the top level targets.json file.
+            /** @var \Tuf\Metadata\TargetsMetadata $targetsMetadata */
+            $targetsMetadata = TargetsMetadata::createFromJson($this->durableStorage['targets.json']);
+            if ($targetsMetadata->hasTarget($target)) {
```
phenaproxima, line 665: Shouldn't this `if` block come _after_ the check if `$targetsMetadata === null`, so that we return as quickly as possible?

src/Client/Updater.php
```diff
@@ -644,6 +640,70 @@ public function download(string $target, ...$extra): PromiseInterface
...
+            $delegatedRoleName = $delegatedRole->getName();
+            if (in_array($delegatedRoleName, $searchedRoles, true)) {
+                // TUF-SPEC-v1.0.9 Section 5.4.5.1
+                // If this role has been visited before, then skip this role (so that cycles in the delegation graph are avoided).
+                continue;
+            }
+            if (!$this->roleDB->roleExists($delegatedRoleName)) {
+                $this->roleDB->addRole($delegatedRole);
+            }
+            if (!$delegatedRole->matchesRolePath($target)) {
```
phenaproxima, line 684: Now that I'm seeing it in use, `matchesRolePath()` is a strange name for this method. It should maybe be something like `matchesPath()`. Follow-up to rename it?

src/Client/Updater.php
```diff
@@ -644,6 +640,70 @@ public function download(string $target, ...$extra): PromiseInterface
...
+                continue;
+            }
+
+            $this->fetchAndVerifyTargetsMetadata($delegatedRoleName);
```
phenaproxima, line 694: Looking at these two lines, it feels like `fetchAndVerifyTargetsMetadata` should return an instance of `TargetsMetadata`.

src/Client/Updater.php
```diff
@@ -644,6 +640,76 @@ public function download(string $target, ...$extra): PromiseInterface
...
+            if (!$this->roleDB->roleExists($delegatedRoleName)) {
+                $this->roleDB->addRole($delegatedRole);
+            }
+            if (!$delegatedRole->matchesRolePath($target)) {
+                // Targets must match the path in all roles in the delegation chain so if the path does not match
+                // do not evaluate this role or any roles it delegates to.
+                // @todo Add fixtures and tests for roles that delegate to other roles with conflicting 'paths' patterns
+                //    to ensure that targets in the nested delegations are not matched in
+                //    https://github.com/php-tuf/php-tuf/issues/142
+                continue;
```
phenaproxima, line 690: Two `continue`s?

phenaproxima (changes requested):

src/Client/Updater.php
tedbow: fixed

src/Client/Updater.php
tedbow: fixed

src/Client/Updater.php
tedbow: Documented this and throwing a UnexpectedValueException if this is not empty and `$targetsMetadata` is empty. This should never be set in a top level call.

src/Client/Updater.php
tedbow: Only the top level `targets.json` needs to be searched here. In other calls we have already called `$newTargetsData->hasTarget($target)` when are looping through delegated and before we call this function recursively.  

If waited to do it here and we only had `$targetsMetadata` we won't have the information to know we should loop through the delegated roles in `$targetsMetadata` based on whether it is `terminating`.  `$targetsMetadata` can't tell us if this is for a terminating role only the metadata file that delegated to this role can tell us. 

You can set up the delegation incorrectly so that you make a terminating delegation but then delegate from that terminating delegation. We need to make sure we don't search those.

I have added a todo to #142 to test for the case.

src/Client/Updater.php
tedbow: good point I will create a PR

src/Client/Updater.php
```diff
@@ -644,6 +643,79 @@ public function download(string $target, ...$extra): PromiseInterface
...
+     * Gets a target metadata object that contains the specified target.
+     *
+     * @param string $target
+     *   The path of the target file. Needs to be known to the most recent
+     *   targets metadata downloaded in ::refresh().
+     * @param \Tuf\Metadata\TargetsMetadata|null $targetsMetadata
+     *   The targets metadata to search or null. If null then the search will
+     *   start at the top level 'targets.json' file.
+     * @param string[] $searchedRoles
+     *   The roles that have already been searched. This is used for recursive
```
phenaproxima, line 656:
```suggestion
     *   The roles that have already been searched. This is for internal use only and should not be passed by calling code.
```

src/Client/Updater.php
```diff
@@ -644,6 +643,79 @@ public function download(string $target, ...$extra): PromiseInterface
...
+     *
+     * @return \Tuf\Metadata\TargetsMetadata|null
+     *   The target metadata with a match for the target, or null no match is
+     *   found.
+     */
+    protected function getMetadataForTarget(string $target, ?TargetsMetadata $targetsMetadata = null, array $searchedRoles = []): ?TargetsMetadata
+    {
+        if ($targetsMetadata === null) {
+            if (!empty($searchedRoles)) {
+                throw new \UnexpectedValueException('$searchedRoles should never be provided by outside calls to \Tuf\Client\Updater::getMetadataForTarget(). It is only used for recursive calls.');
```
phenaproxima, line 667: This feels aggressive. Why not simply empty the array out?

```suggestion
                $searchedRoles = [];
```

phenaproxima (approved):
Couple of minor suggestions, but nothing blocking.

src/Client/Updater.php
tedbow: Because the only reason this would ever happen is if some caller set this because they didn't understand the comment and how this is suppose to work. I don't want to swallow this error because it really means the caller does not understand how this suppose to work.

The should find out this problem rather have it be ignored because their misunderstanding could lead to other problems.

PR #143, opened 2021-04-06
Move verification of downloaded target into new public verify() method

phenaproxima:
Per Seldaek's feedback in composer/composer#9740, it would simplify our integration with Composer if it were possible to verify a downloaded file or string against TUF without relying on TUF to do the downloading too. To that end, let's expose a new `verify()` method which takes a data stream and validates it against a specific known target.


tedbow (approved):
Nice change!

PR #144, opened 2021-04-06
Refactor verifying target metadata file into method

tedbow:
In https://github.com/php-tuf/php-tuf/pull/141 will repeat downloading other delegated target metadata files. This will be the same logic as `targets.json`.  This refactors it into method so we can reuse it.


phenaproxima (approved):
This seems like a pretty straightforward refactoring.

PR #145, opened 2021-04-07
Create Role class

tedbow:
The roleDB is kind of messy. Not only keeps track of roles but also has to know the structure of those roles.

Also `Updater::isKeyIdAcceptableForRole()` could be moved to a new Role class.

In https://github.com/php-tuf/php-tuf/pull/141 we will need more info from special type of role. We could create `DelegatedRole extends Role` with `DelegatedRole::matchTargetPath($target)`

### DIFF
--- a/src/Metadata/ConstraintsTrait.php
+++ b/src/Metadata/ConstraintsTrait.php
@@ -126,4 +126,18 @@ trait ConstraintsTrait
             ],
         ]);
     }
+
+    /**
+     * Gets the role constraints.
+     *
+     * @return \Symfony\Component\Validator\Constraints\Collection
+     *   The role constraints collection.
+     */
+    protected static function getRoleConstraints(): Collection
+    {
+        return new Collection(
+          self::getKeyidsConstraints() +
+          static::getThresholdConstraints()
+        );
+    }
 }
--- a/src/Metadata/ConstraintsTrait.php
+++ b/src/Metadata/ConstraintsTrait.php
@@ -9,12 +9,38 @@ use Symfony\Component\Validator\Constraints\Count;
 use Symfony\Component\Validator\Constraints\GreaterThanOrEqual;
 use Symfony\Component\Validator\Constraints\NotBlank;
 use Symfony\Component\Validator\Constraints\Type;
+use Symfony\Component\Validator\Validation;
+use Tuf\Exception\MetadataException;
 
 /**
  * Trait with methods to provide common constraints.
  */
 trait ConstraintsTrait
 {
+
+    /**
+     * Validates the structure of the metadata.
+     *
+     * @param \ArrayObject $data
+     *   The data to validate.
+     *
+     * @return void
+     *
+     * @throws \Tuf\Exception\MetadataException
+     *    Thrown if validation fails.
+     */
+    protected static function validateWithConstraints(\ArrayObject $data, Collection $collection): void
+    {
+        $validator = Validation::createValidator();
+        $violations = $validator->validate($data, $collection);
+        if (count($violations)) {
+            $exceptionMessages = [];
+            foreach ($violations as $violation) {
+                $exceptionMessages[] = (string) $violation;
+            }
+            throw new MetadataException(implode(",  \n", $exceptionMessages));
+        }
+    }
 
     /**
      * Gets the common hash constraints.
@@ -136,8 +162,8 @@ trait ConstraintsTrait
     protected static function getRoleConstraints(): Collection
     {
         return new Collection(
-          self::getKeyidsConstraints() +
-          static::getThresholdConstraints()
+            self::getKeyidsConstraints() +
+            static::getThresholdConstraints()
         );
     }
 }
--- a/src/Metadata/ConstraintsTrait.php
+++ b/src/Metadata/ConstraintsTrait.php
@@ -23,16 +23,18 @@ trait ConstraintsTrait
      *
      * @param \ArrayObject $data
      *   The data to validate.
+     * @param \Symfony\Component\Validator\Constraints\Collection $constraints
+     *   Th constraints collection for validation.
      *
      * @return void
      *
      * @throws \Tuf\Exception\MetadataException
      *    Thrown if validation fails.
      */
-    protected static function validateWithConstraints(\ArrayObject $data, Collection $collection): void
+    protected static function validate(\ArrayObject $data, Collection $constraints): void
     {
         $validator = Validation::createValidator();
-        $violations = $validator->validate($data, $collection);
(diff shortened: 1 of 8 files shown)


src/RoleDB.php
```diff
@@ -39,32 +39,7 @@ public static function createFromRootMetadata(RootMetadata $rootMetadata, bool $
...
-            */
-
-            // @todo Verify that testing for start of string rather than
-            //     equality has a purpose.
-            if (strncmp($roleName, 'targets', strlen('targets')) === 0) {
-                $roleInfo['paths'] = [];
-                // @todo In the spec, delegations are not part of the role
-                //     structure; delegations reference roles rather than vice
-                //     versa and are in a separate part of the document. Review
-                //     this once https://github.com/php-tuf/php-tuf/issues/52
```
tedbow, line 60: Issue is closed

src/RoleDB.php
```diff
@@ -83,32 +58,20 @@ public function __construct()
...
      *
-     * @throws \Exception
-     *     Thrown if the role already exists.
-     *
-     * @see https://github.com/theupdateframework/specification/blob/v1.0.9/tuf-spec.md#4-document-formats
-     *
-     * @todo Provide more complete documentation of the structure once
-     *     delgation is implemented and fixtures are regenerated. Issues:
-     *     - https://github.com/php-tuf/php-tuf/issues/50
-     *     - https://github.com/php-tuf/php-tuf/issues/52
```
tedbow, line 103: These issue are closed.

src/RoleDB.php
```diff
@@ -131,60 +94,20 @@ public function roleExists(string $roleName)
...
-     * @param string $roleName
-     *    The role name.
-     *
-     * @return integer
-     *     The threshold number of signatures required for the role.
-     *
-     * @throws \Exception
-     *     Thrown if the role does not exist.
-     */
-    public function getRoleThreshold(string $roleName)
```
tedbow, line 182: The method was never actually used.

tedbow (commented):
Just a few notes

src/Role.php
```diff
@@ -0,0 +1,107 @@
...
+     * @param string $keyId
+     *     The key ID to check.
+     *
+     * @return boolean
+     *     TRUE if the key is authorized for the given role, or FALSE
+     *     otherwise.
+     */
+    public function isKeyIdAcceptable($keyId): bool
+    {
+        return in_array($keyId, $this->keyIds);
```
phenaproxima, line 105:
```suggestion
        return in_array($keyId, $this->keyIds, true);
```

src/Role.php
```diff
@@ -0,0 +1,107 @@
...
+     * Checks whether the given key is authorized for the role.
+     *
+     * @param string $keyId
+     *     The key ID to check.
+     *
+     * @return boolean
+     *     TRUE if the key is authorized for the given role, or FALSE
+     *     otherwise.
+     */
+    public function isKeyIdAcceptable($keyId): bool
```
phenaproxima, line 103:
```suggestion
    public function isKeyIdAcceptable(string $keyId): bool
```

src/Role.php
```diff
@@ -0,0 +1,107 @@
...
+     * @param string $name
+     *   The name of the role.
+     *
+     * @return static
+     *
+     * @see @see https://github.com/theupdateframework/specification/blob/v1.0.9/tuf-spec.md#4-document-formats
+     */
+    public static function createFromMetadata(\ArrayObject $roleInfo, string $name): self
+    {
+        return new static(
```
phenaproxima, line 64: Do we want to add some validation?

phenaproxima (changes requested):
Couple minor suggestions, and I'm wondering if this should have validation. It probably needs tests too.

src/Role.php
tedbow: The metadata will always be validate when created so unless some creates their own ArrayObject instance I don't see how it could have bad data.

src/Role.php
tedbow: fixed

src/Role.php
tedbow: fixed

src/Role.php
phenaproxima: I mean...it's pretty easy to create an ArrayObject instance and fill it with bad data. So I'm still +1 for validation, but I'm okay deferring that to a follow-up.

tedbow: Added validation

phenaproxima (changes requested):
Minor requests, but otherwise looks good to me.

src/Metadata/RootMetadata.php
```diff
@@ -30,16 +31,13 @@ protected static function getSignedCollectionOptions(): array
...
-            static::getThresholdConstraints()
-        );
+
         $options['fields']['roles'] = new Collection([
-            'targets' => new Required($roleConstraints),
-            'timestamp' => new Required($roleConstraints),
-            'snapshot' => new Required($roleConstraints),
-            'root' => new Required($roleConstraints),
-            'mirror' => new Optional($roleConstraints),
+            'targets' => new Required(static::getRoleConstraints()),
```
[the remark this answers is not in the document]
tedbow: fixed

src/Metadata/ConstraintsTrait.php
```diff
@@ -9,13 +9,39 @@
...
+     *
+     * @param \ArrayObject $data
+     *   The data to validate.
+     *
+     * @return void
+     *
+     * @throws \Tuf\Exception\MetadataException
+     *    Thrown if validation fails.
+     */
+    protected static function validateWithConstraints(\ArrayObject $data, Collection $collection): void
```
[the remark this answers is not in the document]
tedbow: fixed

src/Role.php
```diff
@@ -0,0 +1,114 @@
...
+     */
+    protected $threshold;
+
+    /**
+     * The key IDs.
+     *
+     * @var array
+     */
+    protected $keyIds;
+
```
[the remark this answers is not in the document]
tedbow: fixed

src/Role.php
```diff
@@ -0,0 +1,114 @@
...
+     * Creates a role object from TUF metadata.
+     *
+     * @param \ArrayObject $roleInfo
+     *   The role information from TUF metadata.
+     * @param string $name
+     *   The name of the role.
+     *
+     * @return static
+     *
+     * @see @see https://github.com/theupdateframework/specification/blob/v1.0.9/tuf-spec.md#4-document-formats
```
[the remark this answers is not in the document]
tedbow: fixed

phenaproxima (approved):

PR #147, opened 2021-04-09
remove phpcodesniffer-composer-installer dependency and unused use statements

tedbow:
We not longer need a direct dependency on `phpcodesniffer-composer-installer` now that we have `php-tuf/phpcodesniffer-standard`. This was causing a requirements error.

Also removing unused use statements because of new `php-tuf/phpcodesniffer-standard` rule.


TravisCarden (approved):
Looks great. (That new sniff is so helpful.) 👍

PR #158, opened 2021-04-13
Metadata changes

SpyroL7:
Changed the few that got away!


tedbow (approved):
@SpyroL7 thanks looks great! I searched for **etaMeta** matching case and found no matches left!
Thanks for the help!

Issue #154, opened 2021-04-12
Incorrect threshold signature computation


joshuagl:
The current implementation of threshold signature computation in `checkSignatures` counts each of multiple signatures with identical authorised keys separately towards the threshold. This allows an attacker with access to a single valid signing key to create multiple valid signatures in order to meet the minimum threshold of keys.

See security advisory https://github.com/theupdateframework/tuf/security/advisories/GHSA-pwqf-9h7j-7mv8 which describes a similar problem in the reference implementation.

pwolanin:
I'll take a look at this

PR #163, opened 2021-04-15
Issue #154 fix for incorrect threshold signature computation

pwolanin:


davidstrauss:
The necessary fixtures to support this work are in #165.

davidstrauss (approved):
Per previous review, this looks good. Glad we're not just fixing but testing on both sides of the fix (attack and happy path).

PR #165, opened 2021-04-15
Fixtures: Threshold = 2 and Related Attack

davidstrauss:
This is to support testing any fix to #154.

### DIFF
--- a/generate_fixtures.py
+++ b/generate_fixtures.py
@@ -186,11 +186,24 @@ class TUFTestFixtureDelegated(TUFTestFixtureSimple):
         self.write_and_publish_repository()
 
 
+class TUFTestFixtureThresholdTwo(TUFTestFixtureBase):
+    def __init__(self):
+        super().__init__()
+        (public_timestamp_key_2, private_timestamp_key_2) = self.write_and_import_keypair(
+            'timestamp_2')
+        self.repository.timestamp.add_verification_key(public_timestamp_key_2)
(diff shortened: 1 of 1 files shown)


generate_fixtures.py
```diff
@@ -186,11 +187,51 @@ def __init__(self):
...
+        (public_timestamp_key_2, private_timestamp_key_2) = self.write_and_import_keypair(
+            'timestamp_2')
+        self.repository.timestamp.add_verification_key(public_timestamp_key_2)
+        self.repository.timestamp.load_signing_key(private_timestamp_key_2)
+        self.repository.timestamp.threshold = 2
+        self.repository.mark_dirty(['timestamp'])
+        self.write_and_publish_repository(export_client=True)
+
+
+class TUFTestFixtureThresholdTwoAttack(TUFTestFixtureBase):
```
pwolanin, line 202: is this supposed to extend TUFTestFixtureThresholdTwo ?

pwolanin: I think extended didn't work as desired - instead I copied some of the code below

davidstrauss: > is this supposed to extend TUFTestFixtureThresholdTwo ?

Sharp eye. I'll fix this.

davidstrauss: I switched it to extend the other class. The output looks correct to me.

pwolanin (approved):
Looks good  - tested locally

Issue #152, opened 2021-04-12
Make sure all methods have return type where possible


tedbow:
We use return types but not consistently. We should use return types whenever possible. 

This might be several PRs depending on the scope of each one. Maybe 1 for `/src` and 1 for `/tests`

ajfwebdev:
I started working on this in [a fork](https://github.com/ajfwebdev/php-tuf). My working branch is feature/152-add-return-types

ajfwebdev:
I submitted a PR with my changes: https://github.com/php-tuf/php-tuf/pull/167/files
The PR was reviewed and merged. Thanks, @tedbow !

tedbow:
#167 fixed this! Thanks @ajfwebdev 

I will create a follow up issue to see if we can have phpcs rule for this. Not sure because we can't add return types if we are implementing an interface. 

We should at least be able to add a rule for the formatting if it exists

PR #167, opened 2021-04-16
Issue 152 Add return type declarations

ajfwebdev:
I went through the whole codebase and added return type declarations where I could.
Also, I changed the style of these declarations to be like: `): int` where there is no space between closing parens and colon, and one space between colon and type.

Issue:
https://github.com/php-tuf/php-tuf/issues/152


tedbow:
Looks great! just a couple things

Found 1 that needs space before colon fix `public static function asNormalizedJson(iterable $structure) : string` 

It looks like `\Tuf\JsonNormalizer::decode()`
`public static function decode(string $json)`

could have same return type as `\Tuf\JsonNormalizer::replaceStdClassWithArrayObject()` because it just returns the value from that method.

`\Tuf\Metadata\MetadataBase::createFromJson()` can be 
`public static function createFromJson(string $json): self()`

there are some others in `/tests`
I used this regex "(protected|public|private)(| static) function .*(?<!__construct)\(.*\)[^\:]" to find these
(I already had this regex to look for these in drupal core)

```
Targets
    Occurrences of '(protected|public|private)(| static) function .*(?<!__construct)\(.*\)[^\:]' in Directory /Users/ted.bowman/projects/php-tuf/tests
Found Occurrences  (11 usages found)
    Unclassified  (11 usages found)
        php-tuf  (11 usages found)
            tests/Client  (3 usages found)
                UpdaterTest.php  (3 usages found)
                    208 public function providerRefreshRepository()
                    316 public function providerRefreshException()
                    502 public function testUpdateRefresh()
            tests/Metadata  (3 usages found)
                RootMetadataTest.php  (3 usages found)
                    83 public function testRequiredRoles(string $missingRole)
                    100 public function providerRequireRoles()
                    131 public function testInvalidRoleName()
            tests/TestHelpers  (1 usage found)
                UtilsTrait.php  (1 usage found)
                    25 public static function getFixturesRealPath(string $fixturesSet, string $path, bool $isDir = true)
```

`provider*()` functions should have `: array`

`test*()` should have `: void`

Thanks

tedbow (changes requested):
Requested some  changes in a comment. thanks

ajfwebdev:
Thanks @tedbow. I've made the suggested changes and updated this PR.

tedbow (approved):
Looks great now! Thanks!!1

Issue #56, opened 2020-09-08
What's the deal with `keyid_hash_algorthims`? Handle and document properly


xjm:
In the process of working on #44 I was comparing our code (which is roughly copied from the Python implementation) to [the spec for the document formats](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#4-document-formats).

According to the code, the `KeyDB` is being created directly out of the root metadata file. However, the spec does not include any mention of `keyid_hash_algorithms`. It seems to be something that's been generated by the Python CLI and is expected by the Python implementation without being in the spec.

In the `1.root.json` fixture, the key data looks like this:
```
  "keys": {
   "044ab30128013448339c79e387b4a47e8aa09e873cd9514763fbeb7610b814af": {
    "keyid_hash_algorithms": [
     "sha256",
     "sha512"
    ],
    "keytype": "ed25519",
    "keyval": {
     "public": "b9bf01ba9353b7271c1ede5ef2c48cccc0823fbcbd516a11f70864920070a326"
    },
    "scheme": "ed25519"
   },
```

I was confused as to why the algorithms were SHA 256 and 512 when we we only support Ed25519. @mbaynton suggested that it's used to create a single checksum hash of the whole key (including the signature _and_ all the associated metadata).

We should figure out why this difference from the spec exists, document it, and maybe submit an issue upstream?

xjm:
Could use feedback from @davidstrauss  as well on this.

davidstrauss:
>  confused as to why the algorithms were SHA 256 and 512 when we we only support Ed25519.

These serve different roles in the design and complement each other; the places where TUF uses SHA are separate from the places where TUF uses Ed25519 (or alternatives like RSA/ECDSA). The SHA-2 hashes (SHA-256 and SHA-512) create secure content digests for validating that content matches a known, trusted digest. An algorithm like Ed25519 supports asymmetric signing algorithm that allows a private keyholder to provide proof that signed content hasn't varied since they signed it.

In simile form: `symmetric encryption : asymmetric encryption :: hashes : signing`

In symmetric encryption and hashes (including HMAC), both authorized parties have access to the same secrets (though there may be no secrets in the hashing case). In asymmetric encryption and signing, only one party has a private key. In a sense, you could sign instead of hash for many cases, but it's unnecessary when you can bundle a hash into another trusted file (which is signed or otherwise assumed to be trustworthy).

[...]

PR #174, opened 2021-04-17
Issue #56 fix use of keyid_hash_algorithms in key data

pwolanin:
The current implementation of generating multiple key IDs and allowing them to be used is not in line with the spec and opens a security hole, per discussion yesterday with @tedbow

### DIFF
--- a/src/KeyDB.php
+++ b/src/KeyDB.php
@@ -79,31 +79,30 @@ class KeyDB
      *     An ArrayAccess object of key metadata. See self::addKey() and the TUF
      *     specification for the array structure.
      *
-     * @return string[]
-     *     An array of hashed key IDs for the key metadata. Each entry is a
-     *     string hash of the key signature and all its associated metadata.
-     *     There is one entry for each hashing algorithm specified in the
-     *     'keyid_hash_algorithms' child array.
+     * @return string
+     *     The hashed key ID for the key metadata using sha256.
      *
      * @see https://github.com/theupdateframework/specification/blob/v1.0.9/tuf-spec.md#4-document-formats
      *
      * @todo https://github.com/php-tuf/php-tuf/issues/56
      */
-    private static function computeKeyIds(\ArrayAccess $keyMeta): array
+    private static function computeKeyId(\ArrayAccess $keyMeta): string
     {
+        // @see https://github.com/secure-systems-lab/securesystemslib/blob/master/securesystemslib/keys.py
+        // The keyid_hash_algorithms array value is based on the TUF settings,
+        // it's not expected to be part of the key data. The fact that it is
+        // currently included is a quirk of the TUF python code that may be
+        // fixed in future versions. For simplicity, hard-code it using the
+        // normal TUF settings.
         $keyCanonicalStruct = [
             'keytype' => $keyMeta['keytype'],
             'scheme' => $keyMeta['scheme'],
-            'keyid_hash_algorithms' => $keyMeta['keyid_hash_algorithms'],
+            'keyid_hash_algorithms' => ['sha256', 'sha512'],
             'keyval' => ['public' => $keyMeta['keyval']['public']],
         ];
         $keyCanonicalForm = JsonNormalizer::asNormalizedJson($keyCanonicalStruct);
 
-        // Generate a hash of the key and its metadata for each of the listed
-        // keyid_hash_algorithms.
-        return array_map(function ($algo) use ($keyCanonicalForm) {
-            return hash($algo, $keyCanonicalForm, false);
-        }, $keyMeta['keyid_hash_algorithms']);
+        return hash('sha256', $keyCanonicalForm, false);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


pwolanin:
This unresolved PR in TUF may date from when multiple key IDs were allowed for the same key.  I'm guessing this was solving by limiting key ID to only be the sha-256 digest in the spec.
https://github.com/theupdateframework/tuf/commit/1a00d87f454e9294325752e838f9fa50c19d48ff

I guess there is still a weird case that 2 keys could use the same public key with different signature schemes, but that would still require them to have both been added to root.json... maybe a risk for delegated keys?

src/KeyDB.php
```diff
@@ -73,37 +74,41 @@ public static function getSupportedKeyTypes(): array
...
+        // currently included is a quirk of the TUF python code that may be
+        // fixed in future versions. Calculate using the normal TUF settings
+        // since this is how it's calculated in the securesystemslib code and
+        // any value for keyid_hash_algorithms in the key data in root.json is
+        // ignored.
         $keyCanonicalStruct = [
             'keytype' => $keyMeta['keytype'],
             'scheme' => $keyMeta['scheme'],
-            'keyid_hash_algorithms' => $keyMeta['keyid_hash_algorithms'],
+            'keyid_hash_algorithms' => ['sha256', 'sha512'],
```
tedbow, line 106: does this come from `securesystemslib.settings.HASH_ALGORITHMS` setting?

If so should this be a setting on our side too? Could securesystemslib setup to use different algorithms or say just `sha256` or just `sha512`?

src/KeyDB.php
```diff
@@ -73,37 +74,41 @@ public static function getSupportedKeyTypes(): array
...
             'keyval' => ['public' => $keyMeta['keyval']['public']],
         ];
         $keyCanonicalForm = JsonNormalizer::asNormalizedJson($keyCanonicalStruct);
 
-        // Generate a hash of the key and its metadata for each of the listed
-        // keyid_hash_algorithms.
-        return array_map(function ($algo) use ($keyCanonicalForm) {
-            return hash($algo, $keyCanonicalForm, false);
-        }, $keyMeta['keyid_hash_algorithms']);
+        return hash('sha256', $keyCanonicalForm, false);
```
tedbow, line 111: Is this the default algo setting that securesystemslib or the python TUF client will use or will this always be `sha256` when creating the server json files?

src/KeyDB.php
pwolanin: It's really coming from the settings the tuf python client sends into the securesystemslib, I think.

If this value ever changed, all keyids would change, so I don't think they would ever change it unless in a future version that's incompatible.  Honestly, it's just stupid that it's even there.

src/KeyDB.php
pwolanin: The TUF client always uses just the sha256 version. That's what's in the spec.

src/KeyDB.php
tedbow: I guess it meant if someone want to set up the python server side. When they are setting it up to use can they change this setting before they are start using it? If so when they need to set up our client differently?

src/KeyDB.php
tedbow: ok. thanks I found this in the spec

> The KEYID of a key is the hexdigest of the SHA-256 hash of the canonical form of the key.

src/KeyDB.php
pwolanin: I don't think they'd ever set it up differently - if they did it would be incompatible with any standard installs of the TUF client.  This quirk is just kind of stupid, so I think have a hard-coded values represents essentially the truth.

src/KeyDB.php
```diff
@@ -39,17 +40,17 @@ class KeyDB
      * @return \Tuf\KeyDB
      *     The constructed key database object.
      *
-     * @throws \Exception
-     *   Thrown if an unsupported key type exists in the metadata.
+     * @throws \Tuf\Exception\InvalidKeyException
+     *   Thrown if an unsupported or invalid key exists in the metadata.
```
ajfwebdev, line 44: Why does the method comment include a `@throws` but the method does not throw an exception?

src/KeyDB.php
tedbow: @pwolanin okay that makes sense. If for some reason it was setup differently our client would fail because the key id computation would not work, correct? I don't think that is bad thing. I just want to make sure  we would fail if the server is not set up as we expect.

src/KeyDB.php
tedbow: I think this ok because `addKey()` is private(for now) so a caller of the public `createFromRootMetadata()` should know the exception could be thrown.

src/KeyDB.php
tedbow: Added a follow up #179 to ensure that we only accept exact expected value for `keyid_hash_algorithms`. This mean if for some reason it was changed on server we would fail.

tedbow (approved):
Looks good! Thanks @pwolanin !

PR #175, opened 2021-04-19
Make Updater::verify() and ::checkLength() protected

phenaproxima:
As per discussion with @tedbow, @effulgentsia, and @xjm, we want to hide the implementation of `Updater::verify()` and `Updater::checkLength()` so that consumers of PHP-TUF that need to use their own download mechanism -- like the Composer integration plugin, for example -- can do so by extending `Updater`, but still take advantage of TUF's verification mechanisms.


tedbow (approved):
Looks good!

PR #177, opened 2021-04-23
61 Added Typed Exception for Role Exists

ajfwebdev:
### Changes
 - 61: Added a new typed exception for when role already exists while adding role.
### Issue
 - https://github.com/php-tuf/php-tuf/issues/61

@tedbow 
This is an MR to see if I'm on the right track for standardizing our exception logic. I'll look for other exceptions to refactor after initial code review of this.

…ding role.

### DIFF
--- a/src/Exception/RoleExistsException.php
+++ b/src/Exception/RoleExistsException.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Tuf\Exception;
+
+/**
+ *  Indicates the role being added already exists.
+ */
+class RoleExistsException extends TufException
+{
+
+}
--- a/src/Exception/RoleExistsException.php
+++ b/src/Exception/RoleExistsException.php
@@ -7,5 +7,4 @@ namespace Tuf\Exception;
  */
 class RoleExistsException extends TufException
 {
-
 }
--- a/src/RoleDB.php
+++ b/src/RoleDB.php
@@ -3,6 +3,7 @@
 namespace Tuf;
 
 use Tuf\Exception\NotFoundException;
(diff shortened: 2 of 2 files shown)


src/Exception/RoleExistsException.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Tuf\Exception;
+
+/**
+ *  Indicates the role being added already exists.
+ */
+class RoleExistsException extends TufException
+{
+
```
phenaproxima, line 10: Minor nit: can you remove this extra blank line?

src/RoleDB.php
```diff
@@ -68,7 +69,7 @@ public function __construct()
     public function addRole(Role $role): void
     {
         if ($this->roleExists($role->getName())) {
-            throw new \Exception('Role already exists: ' . $role->getName());
+            throw new RoleExistsException('Role already exists: ' . $role->getName());
```
phenaproxima, line 72: Looks like we don't have any existing test coverage for RoleDB, which is a shame. Can you open a follow-up issue (no need for a PR yet) to remind us to add test coverage for RoleDB?

phenaproxima (changes requested):
I only request one rinky-dink change and a follow-up issue, and then this is good to go.

src/RoleDB.php
phenaproxima: Filed #178 to address this. :)

phenaproxima (approved):
On second thought, it looks like I can do both of those things myself. So let's not delay getting this in! Approved.

PR #179, opened 2021-04-29
Ensure that keyid_hash_algorithm is equal to exact expected value

tedbow:
In #174  we realized `keyid_hash_algorithms` should not vary and we don't actually need it from metadata to compute the key. Currently in `\Tuf\Metadata\ConstraintsTrait::getKeyConstraints()` we allow any non-empty array of strings. This changes to only except the exact value `["sha256", "sha512"]`

### DIFF
--- a/src/Metadata/ConstraintsTrait.php
+++ b/src/Metadata/ConstraintsTrait.php
@@ -128,6 +128,12 @@ trait ConstraintsTrait
     protected static function getKeyConstraints(): Collection
     {
         return new Collection([
+            // This field is not part of the TUF specification and is being
+            // removed from the Python TUF reference implementation in
+            // https://github.com/theupdateframework/tuf/issues/848.
+            // If it is provided though we only support the default value which
+            // is passed on from a setting in the Python `securesystemslib`
+            // library.
             'keyid_hash_algorithms' => [
                 new Optional(),
                 new EqualTo(['value' =>["sha256", "sha512"]]),
--- a/src/Metadata/ConstraintsTrait.php
+++ b/src/Metadata/ConstraintsTrait.php
@@ -6,6 +6,7 @@ namespace Tuf\Metadata;
 use Symfony\Component\Validator\Constraints\All;
 use Symfony\Component\Validator\Constraints\Collection;
 use Symfony\Component\Validator\Constraints\Count;
+use Symfony\Component\Validator\Constraints\EqualTo;
 use Symfony\Component\Validator\Constraints\GreaterThanOrEqual;
 use Symfony\Component\Validator\Constraints\NotBlank;
 use Symfony\Component\Validator\Constraints\Type;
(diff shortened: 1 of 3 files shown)


src/Metadata/ConstraintsTrait.php
```diff
@@ -127,13 +129,8 @@ protected static function getKeyConstraints(): Collection
...
         return new Collection([
             'keyid_hash_algorithms' => [
-                new Count(['min' => 1]),
-                new Type(['type' => 'array']),
-              // The keys for 'hashes is not know but they all must be strings.
-                new All([
-                    new Type(['type' => 'string']),
-                    new NotBlank(),
-                ]),
+                new Optional(),
```
phenaproxima, line 132: Why are we adding the `Optional` constraint here? If I understand this syntax correctly, previously we expected it to be a non-empty array. Wouldn't making it optional break that?

tedbow: per this issue https://github.com/theupdateframework/tuf/issues/848
`keyid_hash_algorithms` is not going in the JSON metadata files in the future so it should be optional. But currently this data reflects the setting in the `https://github.com/secure-systems-lab/securesystemslib` library that the Python TUF reference implementation uses. Our library(and probably the python library) is built to expect this setup. So if for some reason, which is very unlikely, `securesystemslib` was set up with different hash algorthims we should fail.

phenaproxima: Okay, that makes sense. Can you add a comment explaining why it's optional?

tests/Metadata/RootMetadataTest.php
```diff
@@ -193,4 +192,22 @@ public function testGetRoles(): void
...
+    /**
+     * Test that keyid_hash_algorithms must equal the exact value.
+     *
+     * @see \Tuf\Metadata\ConstraintsTrait::getKeyConstraints()
+     */
+    public function testKeyidHashAlgorithms()
+    {
+        $json = $this->localRepo[$this->validJson];
+        $data = json_decode($json, true);
+        $keyId = array_keys($data['signed']['keys'])[0];
```
phenaproxima, line 205:
```suggestion
        $keyId = key($data['signed']['keys']);
```

src/Metadata/ConstraintsTrait.php
tedbow: Yep just pushed up a comment

tests/Metadata/TargetsMetadataTest.php
```diff
@@ -164,4 +165,22 @@ public function testGetRole(): void
...
+    /**
+     * Test that keyid_hash_algorithms must equal the exact value.
+     *
+     * @see \Tuf\Metadata\ConstraintsTrait::getKeyConstraints()
+     */
+    public function testKeyidHashAlgorithms()
+    {
+        $json = $this->localRepo[$this->validJson];
+        $data = json_decode($json, true);
+        $keyId = array_keys($data['signed']['delegations']['keys'])[0];
```
phenaproxima, line 178:
```suggestion
        $keyId = key($data['signed']['delegations']['keys']);
```

phenaproxima (approved):
Okay, I guess I don't see any problem with this. Just a very minor suggestion, in two places. Approved!

PR #180, opened 2021-05-06
Create a Key class to represent key data.

tedbow:


src/Key.php
```diff
@@ -0,0 +1,131 @@
...
+     * @var string
+     */
+    private $scheme;
+
+    /**
+     * The public key value.
+     *
+     * @var string
+     */
+    private $public;
```
phenaproxima, line 33: 😂😂

src/Metadata/RootMetadata.php
```diff
@@ -70,19 +78,19 @@ public function getRoles(bool $allowUntrustedAccess = false): array
...
+     *   The keys for the metadata.
      */
-    public function getKeys(bool $allowUntrustedAccess = false): \ArrayObject
+    public function getKeys(bool $allowUntrustedAccess = false): array
     {
-        $this->ensureIsTrusted($allowUntrustedAccess);
-        return $this->getSigned()['keys'];
+        if (!isset($this->keys)) {
+            $this->ensureIsTrusted($allowUntrustedAccess);
+            $this->keys = [];
```
phenaproxima, line 88: Is it really necessary to store `$this->keys`? Feels like unnecessary state mutation in this class, which is probably something to be avoided if possible. That said, I don't feel strongly about it.

tests/KeyDBTest.php
```diff
@@ -32,15 +31,12 @@ public function testCreateFromRootMetadata(): void
...
-        // Re-create the the key id which can be used by getKey().
-        $keyNormalized = JsonNormalizer::asNormalizedJson($key);
-        $keyId256 = hash('sha256', $keyNormalized);
-        self::assertSame($jsonEncodedKey, json_encode($keyDb->getKey($keyId256)));
-
-        // Ensure that changing a value in the key does not affect the internal state of the KeyDB object.
-        $key256['keyval']['new_key'] = 'new_value';
-        self::assertSame($jsonEncodedKey, json_encode($keyDb->getKey($keyId256)));
+        $keys = $rootMetadata->getKeys();
+        $key = array_shift($keys);
```
phenaproxima, line 35:
```suggestion
        $key = reset($keys);
```

Nit: it might be wise to avoid state mutation where possible.

tests/Metadata/RootMetadataTest.php
```diff
@@ -40,7 +40,6 @@ public function providerExpectedField(): array
 
         $data[] = ['signed:keys'];
         $firstKey = $this->getFixtureNestedArrayFirstKey('1.root.json', ['signed', 'keys']);
-        $data[] = ["signed:keys:$firstKey:keyid_hash_algorithms"];
```
phenaproxima, line 43: Why was this removed?

tests/Metadata/TargetsMetadataTest.php
```diff
@@ -88,7 +88,6 @@ public function providerExpectedField(): array
         $data[] = ['signed:delegations'];
         $data[] = ['signed:delegations:keys'];
         $firstKey = $this->getFixtureNestedArrayFirstKey($this->validJson, ['signed', 'delegations', 'keys']);
-        $data[] = ["signed:delegations:keys:$firstKey:keyid_hash_algorithms"];
```
phenaproxima, line 91: Same question.

tests/Unit/KeyTest.php
```diff
@@ -0,0 +1,62 @@
...
+
+use Tuf\Key;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @coversDefaultClass \Tuf\Key
+ */
+class KeyTest extends TestCase
+{
+
```
phenaproxima, line 13:
```suggestion
```

tests/Unit/KeyTest.php
```diff
@@ -0,0 +1,62 @@
...
+     * @covers ::createFromMetadata
+     * @covers ::getComputedKeyId
+     * @covers ::getPublic
+     * @covers ::getType
+     *
+     * @dataProvider providerCreateFromMetadata
+     */
+    public function testCreateFromMetadata(array $data): void
+    {
+        $defaultData = [
```
phenaproxima, line 24:
```suggestion
        $data += [
```

tests/Unit/KeyTest.php
```diff
@@ -0,0 +1,62 @@
...
+            'keytype' => 'ed11111',
+            'scheme' => 'scheme-ed11111',
+            'keyval' => new \ArrayObject(['public' => '12345']),
+        ];
+        $data += $defaultData;
```
phenaproxima, line 29:
```suggestion
```

tests/Unit/KeyTest.php
```diff
@@ -0,0 +1,62 @@
...
+        self::assertSame('12345', $key->getPublic());
+        $keySortedCanonicalStruct = [
+            'keyid_hash_algorithms' => ['sha256', 'sha512'],
+            'keytype' => 'ed11111',
+            'keyval' => ['public' => '12345'],
+            'scheme' => 'scheme-ed11111',
+        ];
+        $keyCanonicalForm = json_encode($keySortedCanonicalStruct, JSON_UNESCAPED_SLASHES);
+
+        self::assertSame(hash('sha256', $keyCanonicalForm, false), $key->getComputedKeyId());
```
phenaproxima, line 42:
```suggestion
        self::assertSame(hash('sha256', $keyCanonicalForm), $key->getComputedKeyId());
```

FALSE is the default third argument to `hash()`.

src/Metadata/RootMetadata.php
tedbow: I added this because `RootMetadata` has no setters so you could never change the keys once the object was created was thinking of performance if `getKeys()` was called more than 1x. But right now it is not and this class is not final(maybe it should be) so we don't know a subclass won't add a setter for keys.

Removing

tests/Metadata/RootMetadataTest.php
tedbow: I realized the `\Tuf\Metadata\ConstraintsTrait::getKeyConstraints()` we had the validator for set up wrong. Check out the change there. Basically it should have been optional but it wasn't. It will be removed at some point in the Python library, see #174 and #179.

So this provider was checking that this field was *not* optional. It only passed because `getKeyConstraints()` was wrong. See `\Tuf\Tests\Unit\KeyTest::providerCreateFromMetadata()` that should we can validate now with or without this field.

tests/Metadata/TargetsMetadataTest.php
tedbow: same as above

tests/Unit/KeyTest.php
tedbow: fixed locally

tests/Unit/KeyTest.php
tedbow: fixed

phenaproxima (approved):

PR #186, opened 2021-05-07
Assert only needed delegated json files fetched during refresh and download

tedbow:
In #141 we added nested delegation searching but we didn't have tests to prove the updater does not download all the delegated json files on refresh() or download()


phenaproxima (approved):
@tedbow walked me through this and it makes sense now. Approved!

PR #189, opened 2021-05-10
Add a fixture with a terminating role that still delegates

tedbow:
If a role has `terminating: true` any role it delegates to should not be evaluated. The python library does allow this though

### DIFF
--- a/generate_fixtures.py
+++ b/generate_fixtures.py
@@ -228,16 +228,24 @@ class TUFTestFixtureNestedDelegatedErrors(TUFTestFixtureNestedDelegated):
         level_1_delegation = self.repository.targets._delegated_roles.get('unclaimed')
 
         # Delegate from level_1_delegation to level_1 target-signing key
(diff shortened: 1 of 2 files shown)


generate_fixtures.py
```diff
@@ -228,16 +235,23 @@ def __init__(self):
         level_1_delegation = self.repository.targets._delegated_roles.get('unclaimed')
 
         # Delegate from level_1_delegation to level_1 target-signing key
-        (public_level_2_error_key, private_level_2_error_key) = self.write_and_import_keypair(
+        (public_key, private_key) = self.write_and_import_keypair(
```
phenaproxima, line 238: Why did these variables get renamed?

phenaproxima (approved):
One question but otherwise this makes sense.

generate_fixtures.py
tedbow: I wasn't sure there was point in having separate keys for every role since what are testing if the resources throw a not found exception

tedbow: Reverted this an create a new key for the new delegation.

PR #190, opened 2021-05-11
Update TUF spec implementation to 1.0.10.

phenaproxima:
Based on `https://github.com/theupdateframework/specification/compare/v1.0.9...v1.0.10#diff-b0020aab85257910a5a31f248c201a81cb35e1939545c4b45515e9049e78cdff`.

### DIFF
--- a/src/Client/Updater.php
+++ b/src/Client/Updater.php
@@ -174,59 +174,62 @@ class Updater
 
         $nowDate = $this->getCurrentTime();
 
-        // TUF-SPEC-v1.0.9 Section 5.1.11. Will be used in spec step 5.4.3.
+        // TUF-SPEC-v1.0.10 Section 5.1.11. Will be used in spec step 5.4.?.
         //$consistent = $rootData['consistent'];
 
-        // *TUF-SPEC-v1.0.9 Section 5.2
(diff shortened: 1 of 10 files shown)


phenaproxima:
Ideally I'd like a secondary review from @effulgentsia as well, since this touches TUF spec stuff and it's very possible that I missed something.

src/Client/Updater.php
```diff
@@ -174,59 +174,62 @@ public function refresh(bool $force = false): bool
 
         $nowDate = $this->getCurrentTime();
 
-        // TUF-SPEC-v1.0.9 Section 5.1.11. Will be used in spec step 5.4.3.
+        // TUF-SPEC-v1.0.10 Section 5.1.11. Will be used in spec step 5.4.?.
```
effulgentsia, line 177: 5.4.? is not a step :) 

If we were to add support for consistent snapshots, this would be used in 5.3, 5.4, and 5.5.2. In the meantime, we don't support it and throw an error in supportsConsistentSnapshots(). For clarity, let's have a follow-up to replace this comment and commented-out code line with a @todo to an issue to add support for consistent snapshots.

effulgentsia:
Other than #discussion_r632028269, this looks great to me.

Issue #191, opened 2021-05-13
Support `path_hash_prefixes` from the TUF Spec


tedbow:
Currently we support `paths` for target delegations but not `path_hash_prefixes`. 

This is just placeholder issue so we remember.

I will make a PR shortly to throw a validation error if `path_hash_prefixes` is specified until we support it.

PR #192, opened 2021-05-13
Mark path_hash_prefixes as unsupported

tedbow:
We have #191 to support `path_hash_prefixes` but until then lets throw a validation if it is set

### DIFF
--- a/src/Metadata/TargetsMetadata.php
+++ b/src/Metadata/TargetsMetadata.php
@@ -4,6 +4,7 @@ namespace Tuf\Metadata;
 
 use Symfony\Component\Validator\Constraints\All;
 use Symfony\Component\Validator\Constraints\Collection;
+use Tuf\Constraints\Collection as TufCollection;
 use Symfony\Component\Validator\Constraints\GreaterThanOrEqual;
 use Symfony\Component\Validator\Constraints\NotBlank;
 use Symfony\Component\Validator\Constraints\Optional;
@@ -57,22 +58,27 @@ class TargetsMetadata extends MetadataBase
                 ]),
                 'roles' => new All([
                     new Type('\ArrayObject'),
-                    new Collection([
-                        'name' => [
-                            new NotBlank(),
-                            new Type(['type' => 'string']),
-                        ],
-                        'paths' => [
-                            new Type(['type' => 'array']),
-                            new All([
-                                new Type(['type' => 'string']),
+                    new TufCollection([
+                        'fields' => [
+                            'name' => [
                                 new NotBlank(),
-                            ]),
-                        ],
-                        'terminating' => [
-                            new Type(['type' => 'boolean']),
-                        ],
-                    ] + static::getKeyidsConstraints() + static::getThresholdConstraints()),
+                                new Type(['type' => 'string']),
+                            ],
+                            'paths' => [
+                                new Type(['type' => 'array']),
+                                new All([
+                                    new Type(['type' => 'string']),
+                                    new NotBlank(),
+                                ]),
+                            ],
+                            'terminating' => [
+                                new Type(['type' => 'boolean']),
+                            ],
+                        ] + static::getKeyidsConstraints() + static::getThresholdConstraints(),
+                        // @todo Support 'path_hash_prefixes' in
+                        //    https://github.com/php-tuf/php-tuf/issues/191
+                        'unsupportedFields' => ['path_hash_prefixes'],
+                    ]),
                 ]),
             ]),
         ]);
--- a/tests/Client/UpdaterTest.php
+++ b/tests/Client/UpdaterTest.php
@@ -61,7 +61,7 @@ class UpdaterTest extends TestCase
                 'timestamp' => 2,
                 'snapshot' => 2,
                 'unsupported_target' => null,
-             // We cannot assert the starting versions of 'targets' because it has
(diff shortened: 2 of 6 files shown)


src/Metadata/TargetsMetadata.php
```diff
@@ -57,22 +58,27 @@ protected static function getSignedCollectionOptions(): array
...
-                    new Collection([
-                        'name' => [
-                            new NotBlank(),
-                            new Type(['type' => 'string']),
-                        ],
-                        'paths' => [
-                            new Type(['type' => 'array']),
-                            new All([
-                                new Type(['type' => 'string']),
+                    new TufCollection([
```
phenaproxima, line 61: What's the nature of these changes? It looks like much more has changed here than needs to, but I also don't know this code as well as you.

tests/Client/UpdaterTest.php
```diff
@@ -56,6 +56,14 @@ private static function getFixtureClientStartVersions(string $fixturesSet): arra
                 'targets' => 3,
                 'unclaimed' => 1,
             ],
+            'TUFTestFixtureUnsupportedDelegation' => [
+                'root' => 2,
+                'timestamp' => 2,
+                'snapshot' => 2,
+                'unsupported_target' => null,
+            // We cannot assert the starting versions of 'targets' because it has
```
phenaproxima, line 64:
```suggestion
             // We cannot assert the starting versions of 'targets' because it has
```

tests/Client/UpdaterTest.php
```diff
@@ -745,6 +753,65 @@ public function testUpdateRefresh(): void
...
+            // Ensure that local version of targets has not changed because the
+            // server version is invalid.
+            self::assertSame($this->localRepo['targets.json'], $startingTargets);
+            return;
+        }
+        $this->fail('No exception thrown.');
+    }
+
+    /**
+     * Data provider for testAttackRepoException().
```
phenaproxima, line 795: This doesn't seem accurate...

tests/Client/UpdaterTest.php
```diff
@@ -745,6 +753,65 @@ public function testUpdateRefresh(): void
...
+     * @return array[]
+     *   The test cases.
+     */
+    public function providerUnsupportedRepo(): array
+    {
+        return [
+            [
+                'TUFTestFixtureUnsupportedDelegation',
```
phenaproxima, line 803: Do we...do we actually need this data provider? The test method itself accepts no arguments.

tests/Metadata/UnsupportedFieldsTestTrait.php
```diff
@@ -0,0 +1,43 @@
...
+namespace Tuf\Tests\Metadata;
+
+use Tuf\Exception\MetadataException;
+
+/**
+ * Test method for testing unsupported fields.
+ */
+trait UnsupportedFieldsTestTrait
+{
+
```
phenaproxima, line 12:
```suggestion
```

src/Metadata/TargetsMetadata.php
tedbow: Basically `Symfony\Component\Validator\Constraints\Collection` does support validating a key does **not** exist. So previously we created `\Tuf\Constraints\Collection` to support this an other future add-ons we might want. 

Currently in this code we were using `Symfony\Component\Validator\Constraints\Collection` here because in this part we didn't need exclude and field but now we do so we switch to our extended version.

This allows use to add `'unsupportedFields' => ['path_hash_prefixes'],` and it will validate that for use.

The other change is we now have to nest the previous arguments to the `Collection` constructor under `fields`. This the default option so if don't need to send more than 1 option this is assumed. But now that we want to add another option, `unsupportedFields` we need to nest. This just how default options work in Symfony Validator Constraints.

tests/Client/UpdaterTest.php
tedbow: removed

tests/Client/UpdaterTest.php
tedbow: No, this was left over from when I was doing it differently

tedbow: removed

tests/Client/UpdaterTest.php
```diff
@@ -56,6 +56,14 @@ private static function getFixtureClientStartVersions(string $fixturesSet): arra
                 'targets' => 3,
                 'unclaimed' => 1,
             ],
+            'TUFTestFixtureUnsupportedDelegation' => [
+                'root' => 2,
+                'timestamp' => 2,
+                'snapshot' => 2,
+                'unsupported_target' => null,
+             // We cannot assert the starting versions of 'targets' because it has
```
phenaproxima, line 64:
```suggestion
              // We cannot assert the starting versions of 'targets' because it has
```

src/Metadata/TargetsMetadata.php
phenaproxima: > So previously we created \Tuf\Constraints\Collection to support this an other future add-ons we might want.

Which PR did we do that in?

phenaproxima (approved):

src/Metadata/TargetsMetadata.php
tedbow: https://github.com/php-tuf/php-tuf/pull/99 (via git history)

PR #194, opened 2021-05-13
Update to v1.0.11 of TUF spec

phenaproxima:
This is just a straight search-and-replace, as the spec has no substantive changes in this iteration (see `https://github.com/theupdateframework/specification/compare/v1.0.10...v1.0.11`).


tedbow (approved):
Looks good

PR #196, opened 2021-05-14
Update TUF spec to v1.0.12

phenaproxima:
This PR updates our TUF spec to v1.0.12. As with #194, this is a simple search and replace, because there is no functional change in the spec. (See `https://github.com/theupdateframework/specification/compare/v1.0.11...v1.0.12` for reference.)


tedbow (approved):
looks good!

PR #197, opened 2021-05-14
Update TUF spec to 1.0.13

phenaproxima:
Updates to spec version 1.0.13, according to the differences in `https://github.com/theupdateframework/specification/compare/v1.0.12...v1.0.13`.

The only functional change here was the introduction of a "now+T" time guard to defend against freeze attacks. I discussed this with @tedbow and we decided that the library's default functionality should be unchanged, but that consumers (like the Composer plugin) should be able to provide an adjusted expiration time if they want to, since the spec says that should be left up to the application using TUF.

Beyond that, there was a re-numbering of the spec sections, and that was a little tricky, so I'd ideally like two sign-offs here before that gets merged.

### DIFF
--- a/src/Client/Updater.php
+++ b/src/Client/Updater.php
@@ -88,7 +88,7 @@ class Updater
      *
      *   @return \DateTimeImmutable
      */
-    protected $metadataExpiration;
+    private $metadataExpiration;
 
     /**
      * Updater constructor.
@@ -491,7 +491,7 @@ class Updater
      * @throws \Tuf\Exception\FormatException
      *    Thrown if time format is not valid.
      */
-    protected function getMetadataExpirationTime(): \DateTimeImmutable
+    private function getMetadataExpirationTime(): \DateTimeImmutable
     {
         if (empty($this->metadataExpiration)) {
             throw new \LogicException("::setMetadataExpiration must be called before ::getMetadataExpirationTime");
@@ -764,14 +764,25 @@ class Updater
      *
      * @throws \Tuf\Exception\FormatException
      */
-    protected function setMetadataExpiration()
+    private function setMetadataExpiration()
     {
         if ($this->isRefreshed) {
             throw new \LogicException('::setMetadataExpiration() cannot be called after refresh has been called()');
         } elseif ($this->metadataExpiration) {
             throw new \LogicException('::setMetadataExpiration() should only be called once during ::refresh()');
         }
+        $this->metadataExpiration = $this->createExpirationDate();
+    }
+
+    /**
+     * Creates the expiration date after which metadata will be consider expired.
+     *
+     * @return \DateTimeImmutable
+     *   The expiration date.
+     */
+    protected function createExpirationDate(): \DateTimeImmutable
+    {
         $fakeNow = '2020-01-01T00:00:00Z';
-        $this->metadataExpiration = static::metadataTimestampToDateTime($fakeNow);
+        return static::metadataTimestampToDateTime($fakeNow);
     }
 }
--- a/src/Client/Updater.php
+++ b/src/Client/Updater.php
@@ -336,18 +336,18 @@ class Updater
      *
      * @param \Tuf\Metadata\MetadataBase $metadata
      *     The metadata for the timestamp role.
-     * @param \DateTimeInterface $now
-     *     The current date and time at runtime.
+     * @param \DateTimeInterface $updaterMetadataExpirationTime
+     *     The time after which metadata should be considered expired.
      *
      * @return void
      *
      * @throws FreezeAttackException
      *     Thrown if a potential freeze attack is detected.
      */
-    protected static function checkFreezeAttack(MetadataBase $metadata, \DateTimeInterface $now): void
+    protected static function checkFreezeAttack(MetadataBase $metadata, \DateTimeInterface $updaterMetadataExpirationTime): void
     {
         $metadataExpiration = static::metadataTimestampToDatetime($metadata->getExpires());
-        if ($metadataExpiration < $now) {
+        if ($metadataExpiration < $updaterMetadataExpirationTime) {
             $format = "Remote %s metadata expired on %s";
             throw new FreezeAttackException(sprintf($format, $metadata->getRole(), $metadataExpiration->format('c')));
         }
--- a/src/Client/Updater.php
+++ b/src/Client/Updater.php
@@ -740,7 +740,7 @@ class Updater
     }
 
     /**
-     * Creates the expiration date after which metadata will be consider expired.
+     * Creates the expiration date after which metadata will be considered expired.
      *
      * @return \DateTimeImmutable
      *   The expiration date.
--- a/src/Client/Updater.php
+++ b/src/Client/Updater.php
@@ -748,6 +748,8 @@ class Updater
     protected function createExpirationDate(): \DateTimeImmutable
     {
         $fakeNow = '2020-01-01T00:00:00Z';
-        return static::metadataTimestampToDateTime($fakeNow);
+        // Allow metadata that expires five minutes after ::refresh() starts.
+        $expirationAdditionInterval = \DateInterval::createFromDateString("5 minutes");
+        return static::metadataTimestampToDateTime($fakeNow)->add($expirationAdditionInterval);
     }
 }
--- a/src/Client/Updater.php
+++ b/src/Client/Updater.php
@@ -164,43 +164,46 @@ class Updater
             return true;
         }
 
+        // *TUF-SPEC-v1.0.13 Section 5.0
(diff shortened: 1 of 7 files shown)


src/Client/Updater.php
```diff
@@ -472,14 +475,14 @@ private function updateRoot(RootMetadata &$rootData): void
...
+     * Gets the time beyond which any new metadata will be considered expired.
      *
      * @return \DateTimeImmutable
-     *    The current time.
+     *    A time beyond which any new metadata will be considered expired.
      * @throws \Tuf\Exception\FormatException
      *    Thrown if time format is not valid.
      */
-    private function getCurrentTime(): \DateTimeImmutable
+    protected function getMetadataExpirationTime(): \DateTimeImmutable
```
tedbow, line 485: I don't think we should have this as a function that would call multiple times and it retrieve the time(even if it is fake) each time.

What if we introduced a `protected` `setMetadataExpiration()` that would be called at the very start of `refresh` this would set an a `private $expirationTime`(we could copy our existing logic for getting `$now`.

And then we have a `private` `getMetadataExpirationTime()` that just returns `private $expirationTime` and throws an \`LogicException` if it is called before `setMetadataExpiration()`

That way classes could override `setMetadataExpiration()` but not `getMetadataExpirationTime()` so it would be easy to change how logic it should be before expiration but they couldn't mess with the basic logic.

tedbow (changes requested):

src/Client/Updater.php
```diff
@@ -730,16 +743,35 @@ private function fetchAndVerifyTargetsMetadata(string $role): void
...
+     *
+     * @throws \Tuf\Exception\FormatException
+     */
+    protected function setMetadataExpiration()
+    {
+        if ($this->isRefreshed) {
+            throw new \LogicException('::setMetadataExpiration() cannot be called after refresh has been called()');
+        } elseif ($this->metadataExpiration) {
+            throw new \LogicException('::setMetadataExpiration() should only be called once during ::refresh()');
+        }
```
tedbow, line 773: I put this logic into to avoid this ever being called 2x unless you refresh. I also don't see why you could want to call this outside of refresh.

It thought about  refactoring this into `metadataExpirationCanBeset(): bool` or `validateMetadataExpirationSet()` but that seemed a bit much.

right now to override and retain this logic you have to:

```
protected function setMetadataExpiration()
    {
        parent::setMetadataExpiration();
        $this->metadataExpiration = $this->getMetadataExpirationTime()->add(\DateInterval::createFromDateString('1 day'));
    }
```

src/Client/Updater.php
tedbow: I implemented this but had to make everything  protected for it to overridable.

src/Client/Updater.php
```diff
@@ -159,48 +166,53 @@ public function refresh(bool $force = false): bool
...
+            // *TUF-SPEC-v1.0.13 Section 5.3.2.1 and 5.3.2.2
             $currentStateTimestampData = TimestampMetadata::createFromJson($this->durableStorage['timestamp.json']);
             static::checkRollbackAttack($currentStateTimestampData, $newTimestampData);
         }
 
-        // *TUF-SPEC-v1.0.12 Section 5.2.3
-        static::checkFreezeAttack($newTimestampData, $nowDate);
-        // TUF-SPEC-v1.0.12 Section 5.2.4: Persist timestamp metadata
+        // *TUF-SPEC-v1.0.13 Section 5.3.3
+        static::checkFreezeAttack($newTimestampData, $metadataExpirationTime);
```
tedbow, line 204: For a follow-up I think we don't need to pass in the date here. I think this was done to make it unit test able but I don't think we need to unit test non-public methods.

tedbow (approved):
I did some changes but I think it look good now if @phenaproxima is good with my changes.

src/Client/Updater.php
```diff
@@ -730,16 +743,35 @@ private function fetchAndVerifyTargetsMetadata(string $role): void
...
+     *
+     * This should only be called once during ::refresh() and should not be called
+     * otherwise.
+     *
+     * @throws \Tuf\Exception\FormatException
+     */
+    protected function setMetadataExpiration()
+    {
+        if ($this->isRefreshed) {
+            throw new \LogicException('::setMetadataExpiration() cannot be called after refresh has been called()');
```
phenaproxima, line 770:
```suggestion
            throw new \LogicException('::setMetadataExpiration() cannot be called after refresh has been called');
```

src/Client/Updater.php
```diff
@@ -730,16 +743,35 @@ private function fetchAndVerifyTargetsMetadata(string $role): void
...
+        } elseif ($this->metadataExpiration) {
```
phenaproxima, line 771:
```suggestion
        } elseif (isset($this->metadataExpiration)) {
```
This will allow the expiration to be set to 0, or some other "empty-ish" value, if needed. It's more explicit about what we're really caring about here.

src/Client/Updater.php
phenaproxima: Makes sense; I'm wondering, however, if we really need `getMetadataExpirationTime` to be protected. That might make more sense as a private method. We want to allow subclasses set _set_ an arbitrary expiration time, but once it's set, shouldn't we treat it as an internal aspect of TUF's workings?

src/Client/Updater.php
phenaproxima: One thing I'd prefer is not to force the subclasses to set a specific property; I'd rather that they could return a DateTime object which the updater remembers. So something like:

```
protected function computeMetadataExpiration()
{
  return parent::computeMetadataExpiration()->add(\DateInterval::createFromDateString('1 day'));
}
```

src/Client/Updater.php
tedbow: sounds good. created `createExpirationDate()` as protected and made everything else private

src/Client/Updater.php
tedbow: I pushed up a change with different name. let me know what you think

src/Client/Updater.php
```diff
@@ -83,6 +83,13 @@ class Updater
      */
     protected $isRefreshed = false;
 
+    /**
+     * The time after which metadata should be considered expired.
+     *
+     *   @return \DateTimeImmutable
```
phenaproxima, line 89:
```suggestion
     * @var \DateTimeImmutable
```

phenaproxima (commented):
Minor coding standards thing.

effulgentsia:
This looks good, except it doesn't look like this PR is making any changes to the checkFreezeAttack() function itself, which is currently documented with $now as its 2nd parameter, so that needs to change too.

Also, is there anywhere that we should check if now() is greater than $this->metadataExpiration? I'm not clear from reading the spec on whose responsibility it is to abort an update that is taking longer than what was requested in T. The spec is explicit about needing the metadata to not be expired relative to start_of_update + T, but who is responsible for aborting the update once now() > start_of_update + T?

phenaproxima:
@effulgentsia, my understanding of the spec is that checkFreezeAttack(), in our implementation, should accept an arbitrary time beyond which metadata is considered expired. The variable's name should probably be changed from $now to something clearer, like $expiration. But AFAICT it doesn't need to change its internal implementation, since all it does is check if the metadata's timestamp is less than or equal to a different timestamp, and checkFreezeAttack() has limited visibility.

Am I understanding your question correctly here?

tedbow:
@phenaproxima I think it is confusing because we are basically say T = 0 so we don't have to change our logic. Maybe we should actually add a T value.

phenaproxima:
@tedbow, didn't we agree to leave that aspect of it up to the consumers of TUF (which would be in line with the spec, and what the plugin will probably do once this is merged)?

tedbow:
@phenaproxima but aren't you considering "consumers of TUF " are only libraries that extender `Updater` 

This library should use able without someone extending `Updater`. Right now any direct users would have `T=0`

phenaproxima:
@tedbow I'm open to other suggestions...

tedbow:
@phenaproxima I added minutes 

I also look at the commit that changes this in the spec https://github.com/theupdateframework/specification/commit/e645837a6a47113090d7fcd2a67172b3bf732d25

This is interesting 

> It is further
recommended that the maximum duration is set to the same time period at
which the repository re-signs the timestamp metadata.

So it seem like this value would dependent on how you set the repository. So it seems like would either expect anyone needing to change this based on how often they "re-signs the timestamp metadata" to either
1. Extend `Updater` or
2. We need to expose this setting either via getter or a settings file. We could do this in follow up.

phenaproxima:
Personally, I'd rather make implementors extend `Updater` if they need custom time stretching. Maybe eventually it should be a setting, but IMHO for now it makes more sense to keep it as close to the code level as possible.

PR #201, opened 2021-05-18
Update TUF spec to 1.0.16

phenaproxima:
Update to v1.0.16 of the spec, according to `https://github.com/theupdateframework/specification/compare/v1.0.15...v1.0.16`.

I'm not even mad that it removes the idea of "fixed update expiration time" that we just implemented in #197. I removed the interval stuff and renamed the method to more accurately summarize what it's for.


tedbow (approved):
Looks good!

PR #203, opened 2021-05-19
Fix terminating delegated role logic

tedbow:
initially when I created getMetadataForTarget() I read the spec wrong and thought if the a role is terminating you don't search more of that delegation tree. But actually this abort the search for the metadata. SEe 5.5.6.2.2 in https://github.com/theupdateframework/specification/blob/v1.0.16/tuf-spec.md

This still needs tests


tedbow:
Added the test cases. They should pass but probably need better comments.

phenaproxima (approved):

PR #204, opened 2021-05-20
Move rollback attack checks into metadata classes

phenaproxima:
Updater::checkRollbackAttack() has a bunch of branching logic specific to particular types of metadata. IMHO it would make more sense to move this entire method, and therefore its branching logic, into the particular metadata classes that use it.

### DIFF
--- a/src/Metadata/MetadataAwareMetadataBase.php
+++ b/src/Metadata/MetadataAwareMetadataBase.php
@@ -6,12 +6,21 @@ use Tuf\Exception\MetadataException;
 use Tuf\Exception\PotentialAttackException\RollbackAttackException;
 
 /**
- * Common methods for metadata that contain information about other metadata objects.
+ * Base class for metadata that contains information about other metadata.
  */
-trait MetaFileInfoTrait
+abstract class MetadataAwareMetadataBase extends MetadataBase
 {
+    /**
+     * {@inheritDoc}
+     */
     public function checkRollbackAttack(MetadataBase $remoteMetadata, int $expectedRemoteVersion = null): void
     {
+        parent::checkRollbackAttack($remoteMetadata, $expectedRemoteVersion);
+
+        if (!$remoteMetadata instanceof MetadataAwareMetadataBase) {
+            throw new \InvalidArgumentException();
+        }
+
         $type = $this->getType();
         $localMetaFileInfos = $this->getSigned()['meta'];
         foreach ($localMetaFileInfos as $fileName => $localFileInfo) {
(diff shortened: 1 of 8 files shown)


src/Metadata/MetadataBase.php
```diff
@@ -253,4 +254,41 @@ protected function ensureIsTrusted(bool $allowUntrustedAccess = false): void
...
+     *     The latest metadata fetched from the remote repository.
+     * @param integer|null $expectedRemoteVersion
+     *     If not null this is expected version of remote metadata.
+     *
+     * @return void
+     *
+     * @throws \Tuf\Exception\PotentialAttackException\RollbackAttackException
+     *     Thrown if a potential rollback attack is detected.
+     */
+    public function checkRollbackAttack(MetadataBase $remoteMetadata, int $expectedRemoteVersion = null): void
```
tedbow, line 274: Since this now on `MetadataBase` I am wondering if we should not remove the `$expectedRemoteVersion` parameter here.

The only time we pass it is in `updateRoot()` so it only applies for the Root metadata. We could override this in `RootMetadata` and add the `$expectedRemoteVersion` parameter there

src/Metadata/MetadataAwareMetadataBase.php
```diff
@@ -3,12 +3,37 @@
...
 
 use Tuf\Exception\MetadataException;
+use Tuf\Exception\PotentialAttackException\RollbackAttackException;
 
 /**
- * Common methods for metadata that contain information about other metadata objects.
+ * Base class for metadata that contains information about other metadata.
  */
-trait MetaFileInfoTrait
+abstract class MetadataAwareMetadataBase extends MetadataBase
```
tedbow, line 11: Wonder if we should nave this `FileMetaAwareMetadataBase` because the actual item in the JSON is `meta` and all of the formats are "metadata aware"

tests/Unit/Client/UpdaterTest.php
```diff
@@ -40,20 +41,15 @@ public function testCheckRollbackAttackNoAttack(): void
 
         // The incoming version is newer than the local version, so no
         // rollback attack is present.
-        $localMetadata = $this->getMockBuilder(MetadataBase::class)->disableOriginalConstructor()->getMock();
-        $localMetadata->expects(self::any())->method('getType')->willReturn('any');
-        $localMetadata->expects(self::any())->method('getVersion')->willReturn(1);
+        $localMetadata = $this->mockMetadataFromJson('{"signed": {"_type": "any", "version": 1}}');
```
tedbow, line 44: This test should probably now be in `MetadataBaseTest` with overrides in `TimestampMetadataTest`  and `SnapshotMetadataTest`

tedbow (changes requested):
I like this idea a lot. Asked for a couple changes

src/Metadata/MetadataAwareMetadataBase.php
phenaproxima: Yeah, I wasn't a huge fan of the name either. I don't know how I feel about `FileMeta` though; that's such a strange turn of phrase. I was thinking `MetadataInspectorBase`, but I will bring this up in scrum to see if Travis or Alex have some brilliant idea.

phenaproxima:
This is superseded by #206, so closing it out.

PR #205, opened 2021-05-20
Move section 5.3 into its own method

phenaproxima:
`Updater::refresh()` is a very long method. To make it easier to read through and understand, we should split it up into smaller methods that each handle a coherent piece of the spec. Let's start with section 5.3, which updates timestamp metadata.

This PR doesn't refactor the actual logic in anyway, beyond moving it into its own private method of `Updater`.


tedbow (approved):
Looks good!

PR #207, opened 2021-05-21
Create a factory class for loading trusted metadata

phenaproxima:
This is split off from @tedbow's work in #206. IMHO it makes sense for us to have a metadata factory class, which knows how to load trusted metadata from the persistent storage backend. Later on, we can add a method that also loads the appropriate verifier class for the given role.


tedbow (approved):
Looks good!

PR #208, opened 2021-05-21
Create an explicit interface for durable storage that deals with metadata objects

phenaproxima:
Now that we have things like MetadataFactory, I think it's time we maybe abstracted it further into a storage interface which can read metadata objects from durable storage, and persist trusted metadata objects back to durable storage.


phenaproxima:
I propose the following:

Merge the `ArrayAccess`-based storage with `MetadataFactory` into a new interface: `StorageInterface`. This interface will contain dedicated methods to load and save trusted metadata of each supported type. There will be an abstract base class which implements most of the logic, except for the actual reading and writing.

This will allow us to get rid of things like DurableStorageAccessValidator, which is only there to ensure people aren't using invalid array offsets. With dedicated getters that don't have arguments, and setters which take trusted metadata objects, the need for that kind of validation goes away.

Another aspect of this I like is that calling code does not need to concern itself with the names or file extensions of client-side metadata. You just ask for a type of metadata and you get an object back.

src/Metadata/StorageInterface.php
```diff
@@ -0,0 +1,88 @@
...
+     */
+    public function getRoot(): RootMetadata;
+
+    /**
+     * Persists the trusted root metadata.
+     *
+     * @param \Tuf\Metadata\RootMetadata $metadata
+     *   The root metadata to persist. Must be trusted.
+     */
+    public function saveRoot(RootMetadata $metadata): void;
```
tedbow, line 27: There 4 `save*` methods . `saveTargets` has a different comment for the arg but otherwise there are the same.

Did you think about having just 

`public function saveMetaddata(MetadataBase $metadata): void;`

and then the implementations calling `private/protected` methods based on type?

src/Metadata/StorageInterface.php
```diff
@@ -0,0 +1,88 @@
...
+     *   The targets metadata to persist. Can be for the root `targets` role
+     *   or a delegated role, but must be trusted in any case.
+     */
+    public function saveTargets(TargetsMetadata $metadata): void;
+
+    /**
+     * Deletes stored metadata.
+     *
+     * @param string $name
+     *   The name of the metadata to delete, without file extension.
```
tedbow, line 85: Is this really the file name or the type? like `1.targets.json` would mean you pass `1.targets` or `targets`?

src/Metadata/StorageInterface.php
phenaproxima: I think that's a great idea. Implemented, although I didn't bother with specific protected methods based on type, because that felt like overkill at this stage. We can always refactor that in later if we need it.

src/Metadata/StorageInterface.php
phenaproxima: It's the file name without the extension. You'd pass something like `1.targets` to it. I have clarified in the doc comment.

src/Metadata/StorageBase.php
```diff
@@ -0,0 +1,80 @@
+<?php
+
+namespace Tuf\Metadata;
+
+/**
+ * Defines a base class for saving and loading trusted TUF metadata.
+ */
+abstract class StorageBase implements StorageInterface
```
tedbow, line 8: Do we want to mark this as `@internal`? Since I guess the interface would be the new API?

src/Client/DurableStorage/FileStorage.php
```diff
@@ -2,13 +2,15 @@
...
+use Tuf\Metadata\StorageBase;
+
 /**
  * Defines a simple filesystem-based storage for fetched PHP-TUF metadata.
  *
  * Applications might want to provide an alternative implementation with
  * better performance and error handling.
  */
-class FileStorage implements \ArrayAccess
+class FileStorage extends StorageBase
```
tedbow, line 13: Does composer plugin extend this? If not do we want to take this chance to make this `final` since we are removing public methods and if anyone had extend it before we would break their code

src/Client/Updater.php
```diff
@@ -118,37 +110,18 @@ class Updater
      *       (the actual update data that has been signed).
      *     - confined_target_dirs: (array) @todo What is this for?
      *       https://github.com/php-tuf/php-tuf/issues/161
-     * @param \ArrayAccess $durableStorage
+     *  @param \Tuf\Metadata\StorageInterface $storage
      *     An implementation of \ArrayAccess that stores its contents durably,
      *     as in to disk or a database. Values written for a given repository
      *     should be exposed to future instantiations of the Updater that
      *     interact with the same repository.
```
tedbow, line 117: now out of date

src/Client/Updater.php
```diff
@@ -118,37 +110,18 @@ class Updater
...
-    /**
-     * Gets the type for the file name.
-     *
-     * @param string $fileName
-     *   The file name.
-     *
-     * @return string
-     *   The type.
-     */
-    private static function getFileNameType(string $fileName): string
```
tedbow, line 147: Nice to get rid of this!

src/Client/DurableStorage/FileStorage.php
phenaproxima: It does extend this class, but doesn't override any methods. It just adds helpful static factory methods. I don't mind making this final, but IMHO that should wait until the whole library is in a more final state.

src/Client/Updater.php
```diff
@@ -192,10 +165,10 @@ public function refresh(bool $force = false): bool
 
         // § 5.2
         /** @var \Tuf\Metadata\RootMetadata $rootData */
-        $rootData = $this->metadataFactory->load('root');
+        $rootData = $this->storage->getRoot();
```
tedbow, line 168: So much nicer!

src/Metadata/StorageBase.php
phenaproxima: We could, but since this is a base class I would expect people to extend this. So to me, this is API as well.

That said, I think defining the actual boundaries of the API is probably out of scope for this PR.

src/Client/Updater.php
```diff
@@ -215,7 +188,7 @@ public function refresh(bool $force = false): bool
         $newSnapshotData = SnapshotMetadata::createFromJson($newSnapshotContents);
         $this->universalVerifier->verify(SnapshotMetadata::TYPE, $newSnapshotData);
         // § 5.5.7
-        $this->durableStorage['snapshot.json'] = $newSnapshotContents;
+        $this->storage->save($newSnapshotData);
```
tedbow, line 191: 🎉

src/Client/Updater.php
phenaproxima: Nice catch. Fixed.

src/Client/Updater.php
phenaproxima: It was already not in use, so yeah, good time to remove this wart.

src/Client/DurableStorage/FileStorage.php
tedbow: Maybe we could make a "Make all possible classes final" issue as beta blocker so we don't forget

phenaproxima: Opened #285 for that, along with a milestone for beta.

tedbow (approved):
I was skeptical this was important right now but I have to say this is great improvement!🥇

happy to be proved wrong

PR #209, opened 2021-05-21
Do not allow metadata to be set as untrusted after they are set as trusted

tedbow:
Currently `\Tuf\Metadata\MetadataBase::setIsTrusted()` allows you to mark metadata as trusted or untrusted.

But there is no reason you would keep metadata object around if you knew it was untrusted. All of our calls to this pass true.

We verify and mark as trusted. You should not be able to mark as untrusted. Untrusted is the default state until the metadata is verified.

### DIFF
--- a/src/Metadata/MetadataBase.php
+++ b/src/Metadata/MetadataBase.php
@@ -220,15 +220,8 @@ abstract class MetadataBase
     }
 
     /**
-     * @return boolean
-     *    Whether the metadata is trusted.
-     */
-    public function isTrusted(): bool
-    {
-        return $this->isTrusted;
-    }
-
-    /**
+     * Sets the metadata as trusted.
+     *
      * @return void
      */
     public function setTrusted(): void
@@ -246,7 +239,7 @@ abstract class MetadataBase
      */
     protected function ensureIsTrusted(bool $allowUntrustedAccess = false): void
     {
-        if (!$allowUntrustedAccess && !$this->isTrusted()) {
+        if (!$allowUntrustedAccess && !$this->isTrusted) {
             throw new \RuntimeException("Cannot use untrusted '{$this->getRole()}'. metadata.");
         }
     }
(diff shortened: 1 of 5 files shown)


src/Metadata/MetadataBase.php
```diff
@@ -249,7 +239,7 @@ public function setIsTrusted(bool $isTrusted): void
      */
     protected function ensureIsTrusted(bool $allowUntrustedAccess = false): void
     {
-        if (!$allowUntrustedAccess && !$this->isTrusted()) {
+        if (!$allowUntrustedAccess && !$this->isTrusted) {
```
phenaproxima, line 242: Why did this change?

src/Metadata/MetadataBase.php
```diff
@@ -220,23 +220,13 @@ public function getRole(): string
...
-
-    /**
-     * @param boolean $isTrusted
-     *   Whether the metadata should be trusted.
+     * Sets the metadata as trusted.
      *
      * @return void
      */
-    public function setIsTrusted(bool $isTrusted): void
+    public function setTrusted(): void
```
tedbow, line 227: I am only up for name change since this isn't really a getter but rather more like a 1 way switch. Maybe `markAsTrusted()` ?

phenaproxima (approved):
Seems okay to me, just one question.

src/Metadata/MetadataBase.php
phenaproxima: Maybe just `trust()`?

src/Metadata/MetadataBase.php
tedbow: Since removing the setter I thought it would be good chance remove the getter method also since we don't use it outside of the class.

src/Metadata/MetadataBase.php
tedbow: that works. fixed

phenaproxima (approved):

PR #210, opened 2021-05-24
Import os.path to increase readability of fixture script

phenaproxima:
The fixture generation script is a little hard to read. There are a bunch of reasons for this, but one quick way to help the cause is for us to `from os import path` at the top of the file, so that we can call path functions using `path.foo()` instead of `os.path.foo()`.

### DIFF
--- a/generate_fixtures.py
+++ b/generate_fixtures.py
@@ -1,5 +1,6 @@
 # For instructions on using this script, please see the README.
 
+from os import path
 from tuf import repository_tool as rt
 import os
 import shutil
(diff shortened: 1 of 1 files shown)


phenaproxima:
Apparently this is not considered good practice amongst Python programmers: https://stackoverflow.com/questions/2724348/should-i-use-import-os-path-or-import-os

Since my level of Python experience is very low, it seems wise to defer to best practices on this one. I still would like the fixture generation script to be more readable, but maybe there are other ways to achieve that (or maybe it's already as clean as it could reasonably be, and I'm just unused to the way Python code looks). Closing this out.

PR #211, opened 2021-05-24
Enforce circular role reference ignoring and maximum number of target roles

tedbow:
In `\Tuf\Client\Updater::getMetadataForTarget()` we have `$searchedRoles` but we don't actually add the searched roles to this. 

We also have not implemented this related piece of the spec

>  Otherwise, if an application-specific maximum number of roles have been visited, then go to step 5.6 (so that attackers cannot cause the client to waste excessive bandwidth or time).

Adding this and a test.


tests/Client/UpdaterTest.php
```diff
@@ -288,6 +287,32 @@ public function testVerifiedDelegatedDownload(): void
...
+        $fileName = 'level_1_2_terminating_3_target.txt';
+
+        // Ensure the file can found if the maximum role limit is 100.
+        $this->localRepo = $this->memoryStorageFromFixture($fixturesSet, 'tufclient/tufrepo/metadata/current');
+        $this->testRepo = new TestRepo($fixturesSet);
+        $updater = $this->getSystemInTest($fixturesSet);
+        $testFilePath = static::getFixturesRealPath($fixturesSet, "tufrepo/targets/$fileName", false);
+        $testFileContents = file_get_contents($testFilePath);
+        self::assertNotEmpty($testFileContents);
+        $this->assertSame($testFileContents, $updater->download($fileName)->wait()->getContents());
```
phenaproxima, line 305: Not blocking, but: why do we use `self::assertStuff()` and `$this->assertStuff()` interchangeably? Should we just choose one and stick with it?

phenaproxima (approved):
This makes sense to me.

tests/Client/UpdaterTest.php
tedbow: no reason just  copied different line from different places. will make a PR to clean this up

tedbow: fixed though here

PR #212, opened 2021-05-24
create python function for creating delegations

tedbow:
we know have test fixture where we need to make many delegations. Lets make a function for that

### DIFF
--- a/generate_fixtures.py
+++ b/generate_fixtures.py
@@ -235,13 +235,9 @@ class TUFTestFixtureNestedDelegated(TUFTestFixtureDelegated):
         # are not.
         # See TUFTestFixtureNestedDelegatedErrors
         level_2_terminating_delegation = self.repository.targets._delegated_roles.get('level_2_terminating')
-        (public_level_3_below_terminated_key, private_level_3_below_terminated_key) = self.write_and_import_keypair(
-            'targets_level_3_below_terminated')
-        level_2_terminating_delegation.delegate(
-            'level_3_below_terminated', [public_level_3_below_terminated_key], ['level_1_2_terminating_3_*.txt'])
-        self.write_and_add_target('level_1_2_terminating_3_target.txt', 'level_3_below_terminated')
-        level_2_terminating_delegation('level_3_below_terminated').load_signing_key(
-            private_level_3_below_terminated_key)
+        self.delegate_role(delegator_role=level_2_terminating_delegation, delegated_role_name='level_3_below_terminated',
+                           paths=['level_1_2_terminating_3_*.txt'],
+                           file_name='level_1_2_terminating_3_target.txt')
 
         self.write_and_publish_repository(export_client=False)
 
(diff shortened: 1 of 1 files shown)


tedbow:
We can replace more delegations in generate_fixtures.py but this is a start

generate_fixtures.py
```diff
@@ -76,7 +76,17 @@ def write_and_import_keypair(self, name_dst):
         public_key = rt.import_ed25519_publickey_from_file(pathpub_src)
         private_key = rt.import_ed25519_privatekey_from_file(
             pathpriv_src, password='pw')
-        return (public_key, private_key)
+        return (public_key, private_key)\
+
+    def delegate_role(self, delegator_role, delegated_role_name, paths, file_name):
```
phenaproxima, line 81: I don't think this method should also be able to add a target; things that call this function should create the delegation, then add the target separately.

phenaproxima (approved):
Just one request, otherwise I like it!

generate_fixtures.py
tedbow: This method is able to replace 7 instance where we create a delegation and immediately add a file with the delegation. 

The only place where call `.delegate()` and we can't replace it is in `TUFTestFixtureUnsupportedDelegation` because it is testing that `path_hash_prefixes` is **not** supported. 

I changed the name to `delegate_role_with_file`. I don't see a problem with having this all in 1 method especially since the named parameters make it very clear what the arguments are for.

This not to create an API that others can use but rather just to make our test more readable which I think it does.

tedbow: I updated parameter name from `file_name` to `target_file` to clearer

phenaproxima: > This not to create an API that others can use but rather just to make our test more readable which I think it does.

It's not an API for _others_ to use, that's true ...but it IS part of our internal testing API, much like BrowserTestBase is part of Drupal's. Therefore, I think it _should_ be treated as an important internal API, even if we don't expect others to use it, and we should not let it get sloppy. So I'm still in favor of making them separate calls. I don't really want to block it on this point, but to be honest, I do feel rather strongly about this.

tedbow: They python library we are using already provides `delegate_role()` as `tuf.repository_tool.Targets.delegate()`. `delegate_role_with_file()` is already calls this.

tedbow: Ok I created `delegate_role()`. It is only called 1 time, in `delegate_role_with_file()`.

I don't want to replace the call to `tuf.repository_tool.Targets.delegate()` in `TUFTestFixtureUnsupportedDelegation` because we would have to complicate `delegate_role()` with the extra `path_hash_prefixes` which we only need to prove it is not supported.

Because `delegate_role()` is only called 1x I would rather not create it now and just create it when we need it.  but I don't want to block the issue on it.

phenaproxima (approved):

PR #213, opened 2021-05-25
Update TUF spec to 1.0.17

phenaproxima:
I don't _think_ this had functional changes, but the format of the spec changed and there was a renumbering, so I went through it section by section to confirm the numbering. Hopefully I didn't miss anything, but this should definitely be thoroughly reviewed.

See https://theupdateframework.github.io/specification/v1.0.17.


tedbow (approved):
I went through the client workflow and made sure the number match. I also checked all the commit messages since 1.0.16 https://github.com/theupdateframework/specification/compare/v1.0.16...v1.0.17 and it was all formatting changes

PR #216, opened 2021-05-28
Fix terminating in fixture generation and getMetadataForTarget()

tedbow:
3 problems

1. in #212 we didn't take into account the 1 role `level_2_after_terminating` where terminating should be true. 

`generate_fixtures.TUFTestFixtureBase.delegate_role()`  didn't take this into account.
2.

```
self.delegate_role_with_file(delegator_role=level_1_delegation,
                                     delegated_role_name='level_2_after_terminating',
                                     paths=['level_2_*.txt'],
                                     target_file='level_2_after_terminating_unfindable.txt')
```
The paths here 'level_2_*.txt' will never have a match for the role was delegated from 'unclaimed' which has a patch 'level_1_*.txt'.

File must match all level of the chain so it will not be found because of that not because it is after the delegation.

```
if (!$delegatedRole->matchesPath($target)) {
     // Targets must match the path in all roles in the delegation chain so if the path does not match
     // do not evaluate this role or any roles it delegates to.
     continue;
 }

 $this->fetchAndVerifyTargetsMetadata($delegatedRoleName);
 /** @var \Tuf\Metadata\TargetsMetadata $newTargetsData */
 $newTargetsData = $this->metadataFactory->load($delegatedRoleName);
 if ($newTargetsData->hasTarget($target)) {
     return $newTargetsData;
 }
 $searchedRoles[] = $delegatedRole;
 // § 5.6.7.2.1
 //  If the current delegation is a multi-role delegation, recursively visit each role, and check that each has signed exactly the same non-custom metadata (i.e., length and hashes) about the target (or the lack of any such metadata).
 if ($matchingTargetMetadata = $this->getMetadataForTarget($target, $newTargetsData, $searchedRoles)) {
     return $matchingTargetMetadata;
 }
 if ($delegatedRole->isTerminating()) {
     // § 5.6.7.2.2
     // If the role is terminating then abort searching for a target.
     return null;
 }          
```

This is part of the logic for evaluating each delegate role.

At the top check if the path matches if it does not we go to the next delegated role. 
At the bottom we check if the role is terminating and return null if it is.

So if we have terminating role that does not match target path its termination will not take affect. I think this is wrong but we should check.


src/Client/Updater.php
```diff
@@ -518,24 +518,24 @@ protected function getMetadataForTarget(string $target, ?TargetsMetadata $target
             }
 
             $this->signatureVerifier->addRole($delegatedRole);
-            if (!$delegatedRole->matchesPath($target)) {
+            if ($delegatedRole->matchesPath($target)) {
                 // Targets must match the path in all roles in the delegation chain so if the path does not match
```
phenaproxima, line 522:
```suggestion
                // Targets must match the path in all roles in the delegation chain, so if the path does not match,
```

phenaproxima (approved):
This makes sense to me. One punctuation nitpick.

tedbow:
I checked the last TUF spec to make sure we don't have to change the logic again.

The description has changed

See latest 1.0.19 https://theupdateframework.github.io/specification/v1.0.19/#update-targets

and 1.0.17 the version we are on: https://theupdateframework.github.io/specification/v1.0.17/#update-targets

This switch the order that `terminating` is evaluated. So we should figure out how to get to 1.0.19 and do that version

tedbow:
See this slack convo https://cloud-native.slack.com/archives/C8NMD3QJ3/p1622232421026200

> so if the paths don’t match then you just skip and don’t care if is terminating(because it is not evaluated) correct?
@trishankatdatadog confirm this is correct

a simplified example https://github.com/trishankatdatadog/tuf-on-a-plane/blob/ee8049a63ea439d72509f6e876eef08fb50c6ab1/src/tuf_on_a_plane/repository.py#L319-L403

tedbow:
Making this a draft because we still need clarification on how this should work

tedbow:
After chatting more with @trishankatdatadog I think 1 & 2 from my original comment on the issue are still a problem but 3 is not.

phenaproxima (approved):

src/Client/Updater.php
phenaproxima: @tedbow Thoughts on this change?

fixtures/TUFTestFixtureTerminatingDelegation/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+        .create_target('b.txt', signing_role='b') \
+        .delegate('c', ['*.txt'], parent='b') \
+        .create_target('c.txt', signing_role='c') \
+        .delegate('d', ['*.txt'], parent='b') \
+        .create_target('d.txt', signing_role='d') \
+        .delegate('e', ['*.txt'], parent='a') \
+        .create_target('e.txt', signing_role='e') \
+        .delegate('f', ['*.txt']) \
+        .create_target('f.txt', signing_role='f') \
+        .publish()
```
tedbow, line 20: @phenaproxima I haven't made the phpunit test that uses this fixture yet but I think this is how to make the test case specified in https://github.com/theupdateframework/specification/issues/168

I looked through the json files and I think it produces the correct fixture but let me know if you think it produces the correct fixture.

BTW I love how easy this is to make with the new `FixtureBuilder` class !

src/Client/Updater.php
```diff
@@ -518,28 +518,26 @@ protected function getMetadataForTarget(string $target, ?TargetsMetadata $target
...
-                return null;
+                $this->fetchAndVerifyTargetsMetadata($delegatedRoleName);
+                /** @var \Tuf\Metadata\TargetsMetadata $newTargetsData */
+                $newTargetsData = $this->metadataFactory->load($delegatedRoleName);
+                if ($newTargetsData->hasTarget($target)) {
+                    return $newTargetsData;
+                }
+                $searchedRoles[] = $delegatedRoleName;
+                // § 5.6.7.2.1
+                //  If the current delegation is a multi-role delegation, recursively visit each role, and check that each has signed exactly the same non-custom metadata (i.e., length and hashes) about the target (or the lack of any such metadata).
```
phenaproxima, line 532: Question: how do we know if the current delegation is a "multi-role" delegation? What does that mean?

tests/Client/UpdaterTest.php
```diff
@@ -277,6 +277,20 @@ public function testVerifiedDelegatedDownload(): void
                 'level_3' => 1,
                 'level_3_below_terminated' => 1,
             ],
+            // Roles after terminating delegation where the target path does not
```
phenaproxima, line 280: "...where the target path does not the terminating role are..." seems like odd phrasing. Missing a word?

phenaproxima (changes requested):

src/Client/Updater.php
tedbow: "multi-role" wording was wrong. This was taken out of spec. I thought it meant a role that delegate to more than 1 more role. but it actually refers to something else that is not in 1.x version of the spec and is being considered for 2.x.

I updated the wording from the current version of the spec.

tests/Client/UpdaterTest.php
tedbow: Added more context. Take another look

src/Metadata/Factory.php
```diff
@@ -50,7 +50,7 @@ public function load(string $role): ?MetadataBase
                     $currentMetadata = TimestampMetadata::createFromJson($json);
                     break;
                 default:
-                    $currentMetadata = TargetsMetadata::createFromJson($json);
+                    $currentMetadata = TargetsMetadata::createFromJson($json, $role);
```
tedbow, line 53: This is the only unrelated fix. We probably need test coverage for the `Factory` class in another PR. 

the 'targets' type is the only  type where role doesn't always equal type. 

without this fix `$delegatedTargetsMetadata->getRole()` would always equal 'targets' where it should equal the delegated role name.

src/Client/Updater.php
```diff
@@ -471,76 +471,26 @@ public function download(string $target, ...$extra): PromiseInterface
...
-            //  If the current delegation is a multi-role delegation, recursively visit each role, and check that each has signed exactly the same non-custom metadata (i.e., length and hashes) about the target (or the lack of any such metadata).
-            if ($matchingTargetMetadata = $this->getMetadataForTarget($target, $newTargetsData, $searchedRoles)) {
-                return $matchingTargetMetadata;
-            }
-            if ($delegatedRole->isTerminating()) {
-                // § 5.6.7.2.2
-                // If the role is terminating then abort searching for a target.
-                return null;
-            }
+        $searchResultMetadata = $this->performDelegationSearch($targetsMetadata, $target, ['targets']);
```
phenaproxima, line 490: IMHO, something like `searchDelegatedRolesForTarget()` would be a clearer method name.

src/Client/Updater.php
```diff
@@ -575,4 +525,70 @@ private function getUpdateStartTime(): \DateTimeImmutable
     {
         return (new \DateTimeImmutable())->setTimestamp($this->clock->getCurrentTime());
     }
+
+    /**
+     * Performs a delegation search for a target metadata object.
```
phenaproxima, line 530: I'd rephrase this a bit: maybe "Searches delegated roles for metadata concerning a specific target"?

src/Client/Updater.php
```diff
@@ -575,4 +525,70 @@ private function getUpdateStartTime(): \DateTimeImmutable
...
+     *
+     * @param \Tuf\Metadata\TargetsMetadata|null $targetsMetadata
+     *   The targets metadata to search.
+     * @param string $target
+     *   The path of the target file. Needs to be known to the most recent
+     *   targets metadata downloaded in ::refresh().
+     * @param string[] $searchedRoles
+     *   The roles that have already been searched. This is for internal use only and should not be passed by calling code.
+     *   calls to this function and should be provided by any callers.
```
phenaproxima, line 539:
```suggestion
```

src/Client/Updater.php
```diff
@@ -575,4 +525,70 @@ private function getUpdateStartTime(): \DateTimeImmutable
...
+     * @param bool $terminated
```
phenaproxima, line 540: No description of this one? At the very least it should say that it's for internal use only.

src/Client/Updater.php
```diff
@@ -575,4 +525,70 @@ private function getUpdateStartTime(): \DateTimeImmutable
...
+     *
+     * @return \Tuf\Metadata\TargetsMetadata|null
+     *   The target metadata that is the end node for search or null if the target metadata does not make any delegations.
```
phenaproxima, line 543: I'd rephrase this too. Maybe "The targets metadata that was the last visited node in the delegation tree, or null if there were no delegated roles to search. If targets metadata is returned, it may or may not contain information about the desired target."

src/Client/Updater.php
```diff
@@ -575,4 +525,70 @@ private function getUpdateStartTime(): \DateTimeImmutable
...
+                // If this role has been visited before, then skip this role (so that cycles in the delegation graph are avoided).
+                continue;
+            }
+            if (count($searchedRoles) > static::MAXIMUM_TARGET_ROLES) {
+                return $delegatedTargetsMetadata ?? null;
+            }
+
+            $this->signatureVerifier->addRole($delegatedRole);
+            if ($delegatedRole->matchesPath($target)) {
+                // Targets must match the path in all roles in the delegation chain, so if the path does not match,
```
phenaproxima, line 564:
```suggestion
                // Targets must match the paths in of all roles in the delegation chain, so if the path does not match,
```

src/Client/Updater.php
```diff
@@ -575,4 +525,70 @@ private function getUpdateStartTime(): \DateTimeImmutable
...
+                $searchedRoles[] = $delegatedRoleName;
+                // § 5.6.7.2.1
+                // Recursively search the list of delegations in order of appearance.
+                if ($resolvedTargetMetadata = $this->performDelegationSearch($delegatedTargetsMetadata, $target, $searchedRoles, $terminated)) {
+                    if ($terminated || $resolvedTargetMetadata->hasTarget($target)) {
+                        return $resolvedTargetMetadata;
+                    }
+                }
+
+                if ($delegatedRole->isTerminating()) {
```
phenaproxima, line 585:
```suggestion
                if ($terminated) {
```

tests/Client/UpdaterTest.php
```diff
@@ -314,6 +315,262 @@ public function providerVerifiedDelegatedDownload(): array
                     'level_3_below_terminated' => 1,
                 ],
             ],
+            // A terminating role only has an effect if the target path matches
+            // the role, otherwise the role is not evaluated.
+            // Roles after terminating delegation where the target path does match not
```
phenaproxima, line 320: "After" is an ambiguous term here. Would "below" maybe be more accurate?

tests/Client/UpdaterTest.php
```diff
@@ -314,6 +315,262 @@ public function providerVerifiedDelegatedDownload(): array
...
+            // the terminating role are evaluated.
```
phenaproxima, line 321: "...where the target path does not match ARE evaluated"? Shouldn't this be "are not"?

tests/Client/UpdaterTest.php
```diff
@@ -314,6 +315,262 @@ public function providerVerifiedDelegatedDownload(): array
...
+                    'targets' => 2,
+                    'a' => 1,
+                    'b' => 1,
+                    'c' => 1,
+                    'd' => 1,
+                    'e' => null,
+                    'f' => null,
+                ],
+            ],
+            'TUFTestFixture3LevelDelegation e.txt' => [
```
phenaproxima, line 542: It'd be nice if all these cases had some comments above them explaining the expected outcome.

phenaproxima (changes requested):
This is really good. I'm super happy about the robust test coverage. My concerns lie more with overall grokkability -- the TUF spec, as we know, is poorly worded about this recursive search for targets within a delegation tree, and therefore I think we need to be completely clear about what we're doing and how we expect it to work. I think my biggest wishlist item after this review is inline comments for all the test cases where, at the very least, we explain what the expected outcome is.

src/Client/Updater.php
tedbow: This actually will cause tests to fail. 

If you look at the case `TUFTestFixtureTerminatingDelegation f.txt` this will fail because F will be found when it should not be found. Same case with `TUFTestFixtureNestedTerminatingNonDelegatingDelegation d.txt` this should throw a not found.

We actually don't need 

```
 if ($delegatedRole->isTerminating()) {
       $terminated = true;
 }
```
Before our recursive call to `performDelegationSearch()`. I moved  it after.

I can update comments in code later to explain why this is the case. Sorry don't have time right now.

phenaproxima: Okay, no problem. I agree that more comments will be useful.

src/Client/Updater.php
tedbow: Updated

src/Client/Updater.php
tedbow: I updated this logic to make it simpler. Take another look

src/Client/Updater.php
```diff
@@ -471,78 +471,23 @@ public function download(string $target, ...$extra): PromiseInterface
...
-     *   start at the top level 'targets.json' file.
-     * @param string[] $searchedRoles
-     *   The roles that have already been searched. This is for internal use only and should not be passed by calling code.
-     *   calls to this function and should be provided by any callers.
+     *   The path of the target file.
      *
      * @return \Tuf\Metadata\TargetsMetadata|null
-     *   The target metadata with a match for the target, or null no match is
-     *   found.
+     *   The target metadata with a match for the target or null no match is found.
```
phenaproxima, line 480:
```suggestion
     *   The target metadata with a match for the target or null if no match is found.
```

src/Client/Updater.php
```diff
@@ -575,4 +520,67 @@ private function getUpdateStartTime(): \DateTimeImmutable
     {
         return (new \DateTimeImmutable())->setTimestamp($this->clock->getCurrentTime());
     }
+
+    /**
+     * Performs a search of delegate roles for a metadata containing information about a target.
```
phenaproxima, line 526:
```suggestion
     * Searches delegated roles for metadata containing information about a target.
```

src/Client/Updater.php
tedbow: added description

src/Client/Updater.php
```diff
@@ -575,4 +520,67 @@ private function getUpdateStartTime(): \DateTimeImmutable
...
+    private function searchDelegatedRolesForTarget(TargetsMetadata $targetsMetadata, string $target, array $searchedRoles, bool &$terminated = false): ?TargetsMetadata
+    {
+        foreach ($targetsMetadata->getDelegatedKeys() as $keyId => $delegatedKey) {
+            $this->signatureVerifier->addKey($keyId, $delegatedKey);
+        }
+        foreach ($targetsMetadata->getDelegatedRoles() as $delegatedRole) {
+            $delegatedRoleName = $delegatedRole->getName();
+            if (in_array($delegatedRoleName, $searchedRoles, true)) {
+                // § 5.6.7.1
+                // If this role has been visited before, then skip this role (so that cycles in the delegation graph are avoided).
```
phenaproxima, line 552:
```suggestion
                // If this role has been visited before, skip it (to avoid cycles in the delegation graph).
```

tests/Client/UpdaterTest.php
```diff
@@ -242,6 +242,7 @@ public function testVerifiedDelegatedDownload(string $fixturesSet, string $deleg
     public function providerVerifiedDelegatedDownload(): array
     {
         return [
+          // Test cases using the TUFTestFixtureNestedDelegated fixture
```
phenaproxima, line 245:
```suggestion
           // Test cases using the TUFTestFixtureNestedDelegated fixture
```

phenaproxima (changes requested):

src/Client/Updater.php
tedbow: @phenaproxima fyi part of the simplification here is that `searchDelegatedRolesForTarget()` only returns either the metadata contains the target info or null if no result. Before it could also return last metadata searched regardless of whether the target was found.

src/Client/Updater.php
tedbow: fixed

src/Client/Updater.php
tedbow: I simplified this method so only returns a targets metadata object if it has the info about the file. So the comment is simpler now

src/Client/Updater.php
tedbow: fixed

tests/Client/UpdaterTest.php
tedbow: No "below" to be would mean roles the role with the terminating delegation delegated to it self.

If you look at this tree

```
#
#             Targets
#             /     \
#            a       f
#           / \
#          b   e
#         / \
#        c   d
#
```
If we are talking about "b" I would consider "c" and "d" "below". I would consider "e" after because it would be evaluated after b and it's delegations.

tests/Client/UpdaterTest.php
tedbow: fixed

tedbow:
> I think my biggest wishlist item after this review is inline comments for all the test cases where, at the very least, we explain what the expected outcome is.

I wondering if the test cases would be easier to understand if we combined `testVerifiedDelegatedDownload()` and `testDelegationErrors()` this could be another PR but....

Right now `testDelegationErrors()` might be poorly named. It is basically when you can't find target for some reason like a terminating delegation before the targets file that has the info is evaluated.

So basically these 2 test methods maybe be better named `testTargetFound()` and `testTargetNotFound()`

But regardless of the naming if we wanted to see all the test cases for 1 fixture you would have to look at 2 provider methods.
So for example if you want to see all the test cases for `TUFTestFixtureTerminatingDelegation` now you would have to look at `providerVerifiedDelegatedDownload()` for targets `targets.txt a.txt b.txt c.txt d.txt` and `providerDelegationErrors()` for `e.txt f.txt`.

We could combine this into 1 test method `testTargetDownload()` with 1 provider and add extra param `bool $expectFound`.
That way you could see 1 block of tests cases all the found and not found target.

src/Client/Updater.php
```diff
@@ -471,78 +471,24 @@ public function download(string $target, ...$extra): PromiseInterface
...
-     *   start at the top level 'targets.json' file.
-     * @param string[] $searchedRoles
-     *   The roles that have already been searched. This is for internal use only and should not be passed by calling code.
-     *   calls to this function and should be provided by any callers.
+     *   The path of the target file.
      *
      * @return \Tuf\Metadata\TargetsMetadata|null
-     *   The target metadata with a match for the target, or null no match is
-     *   found.
+     *   The target metadata with a match for the target or null if no match is found.
```
phenaproxima, line 480:
```suggestion
     *   The targets metadata with information about the desired target, or null if no relevant metadata is found.
```

src/Client/Updater.php
```diff
@@ -575,4 +521,68 @@ private function getUpdateStartTime(): \DateTimeImmutable
...
+                // § 5.6.7.1
+                // If this role has been visited before, skip it (to avoid cycles in the delegation graph).
+                continue;
+            }
+            if (count($searchedRoles) > static::MAXIMUM_TARGET_ROLES) {
+                return null;
+            }
+
+            $this->signatureVerifier->addRole($delegatedRole);
+            // Targets must match the paths in of all roles in the delegation chain, so if the path does not match,
```
phenaproxima, line 559:
```suggestion
            // Targets must match the paths of all roles in the delegation chain, so if the path does not match,
```

tests/Client/UpdaterTest.php
phenaproxima: Yeah, I agree with that interpretation. To make this clearer in the comment, what if we said "after (i.e., next to)"?

tests/Client/UpdaterTest.php
```diff
@@ -436,8 +692,8 @@ public function providerDelegationErrors(): array
                     'timestamp' => 6,
                     'snapshot' => 6,
                     'targets' => 6,
-                    // The client does not update the 'unclaimed.json' file because
-                    // the target file does not match the 'paths' property for the role.
+                // The client does not update the 'unclaimed.json' file because
```
phenaproxima, line 695: Should this indentation have been changed?

tests/Client/UpdaterTest.php
```diff
@@ -446,33 +702,160 @@ public function providerDelegationErrors(): array
...
+                    'b' => 1,
+                    'c' => 1,
+                    'd' => 1,
+                    'e' => 1,
+                    'f' => 1,
+                ],
+            ],
+            // A search for non existent target that does match the paths
+            // should not search any of the tree.
+            'TUFTestFixture3LevelDelegation z.zip' => [
```
phenaproxima, line 843: I think that we should change these keys to actually describe the expected outcome. For example, this one could maybe be "non-existent target, does not match paths, tree not searched" or something like that. Just naming these data sets after the fixture and target name, is simply repeating arguments that are already in the data set.

phenaproxima (changes requested):

tests/Client/UpdaterTest.php
phenaproxima: I'd be okay doing this in a follow-up.

tests/Client/UpdaterTest.php
phenaproxima: Changed it.

phenaproxima (approved):
Made a small change, but otherwise I think I'm okay deferring documentation to a follow-up.

PR #217, opened 2021-06-02
Add known security vulnerabilities check on CI

TravisCarden:
This adds a check for dependencies for known security vulnerabilities to CI builds.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -134,6 +134,9 @@ jobs:
       - name: Run test suite
         run: composer test
 
+      - name: Check dependencies for known security vulnerabilities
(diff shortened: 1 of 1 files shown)


phenaproxima (approved):
No objections, but I'd like @tedbow to cross-check before we merge this.

phenaproxima:
Since Composer now has an `audit` command, we should probably use that instead.

phenaproxima (changes requested):
Could we use `composer audit` for this instead?

TravisCarden:
Sure, @phenaproxima! Let's just do what we did in https://github.com/php-tuf/composer-integration/pull/62.

PR #218, opened 2021-06-04
Update TUF spec to 1.0.18

phenaproxima:
See `https://github.com/theupdateframework/specification/compare/v1.0.17...v1.0.18` for the diff. Doesn't look like there were any functional changes or re-numberings to worry about.


tedbow (approved):
looks good!

PR #220, opened 2021-06-10
Regenerate and commit all test fixtures

phenaproxima:
After discussion with @tedbow, we've decided to develop the conformance suite in here with the eventual goal of splitting them out into their own package (php-tuf/conformance). As a first step, we need to commit our existing fixtures so we can track changes to them. This also means we no longer need to build the fixtures every time on GitHub Actions.


tedbow (approved):
looks good!

PR #221, opened 2021-06-10
Do not immediately publish repositories after initialization in generate_fixtures.py

phenaproxima:
Right now, when building fixtures, we publish the repo immediately after initializing the basic roles. This adds wasteful extra overhead and doesn't appear to be necessary for anything. Let's remove that.


tedbow (approved):
Looks great!

PR #223, opened 2021-06-10
Rename fixture subdirectories to "client" and "server" for clarity

phenaproxima:
Right now, the naming of these directories doesn't make it super clear which one contains the server-side metadata, and which one contains the client-side metadata. Let's rename these so it's clearer.


tedbow (approved):
Looks good!

PR #224, opened 2021-06-11
Create FixtureBuilder class and convert TUFTestFixtureSimple to use it

phenaproxima:
Pretty much what it says. This brings in the FixtureBuilder from php-tuf/conformance and uses it to generate TUFTestFixtureSimple. This also moves *all* fixture-related stuff, including the Pipfile, into the `fixtures` directory.


fixtures/build-all.py
```diff
@@ -0,0 +1,9 @@
+import TUFTestFixtureSimple
+
+from unittest import mock
+
+@mock.patch('time.time', mock.MagicMock(return_value=1577836800))
+def build_all():
+    TUFTestFixtureSimple.build()
+
+build_all()
```
tedbow, line 9: Do we actually need this new file? 
Can't we just update `generate_fixtures.py`

1. Delete the line that calls TUFTestFixtureSimple()
2. replace it with TUFTestFixtureSimple.build()

Once we replace all the fixture generation  `generate_fixtures.py` would only have TUFTestFixture*.build() calls and all class could be removed.
Otherwise right now the readme is not accurate as far as generating fixtures and it would be a 2 step process.

fixtures/Pipfile
```diff
@@ -0,0 +1,27 @@
...
+requests = "==2.24.0"
+securesystemslib = "==0.20.1"
+six = "==1.15.0"
+tuf = "==0.14.0"
+urllib3 = "==1.25.10"
+PyNaCl = "==1.4.0"
+flake8 = "*"
+
+[requires]
+python_version = "3.9"
```
tedbow, line 27: If we just updating  `generate_fixtures.py` would need a new Pipfile?

While I hope we get all the fixture generation converted to use `FixtureBuilder` quickly if we run into problem and it takes longer I would rather not have the repo be in this confusing state of having 2 steps create fixtures.

Related to that I wonder how hard it would be create a github actions to run  `generate_fixtures.py` and confirm that our fixtures do not change. Because anytime you update the generation of fixtures that change should be committed and we want to enforce that.

tedbow (changes requested):

fixtures/build-all.py
tedbow: @phenaproxima  to save time pushed a changed removing `fixtures/build-all.py` for now. I am calling `fixtures.TUFTestFixtureSimple.build()` from  `generate_fixtures.py`. This way all the fixtures are generated from 1 file as documented in the readme.

Once we have converted all fixtures we can either delete `generate_fixtures.py` and add back `fixtures/build-all.py` or just leave `generate_fixtures.py` with all the new builders.

Let me if this works for you.

tedbow (approved):
I am good with as it stands without the 2 scripts to build fixtures. @phenaproxima if that doesn't work for you lets chat

PR #226, opened 2021-06-11
Update Python TUF and securesystemslib

phenaproxima:
This should help us fix problems with fixtures being generated non-deterministically. Updating securesystemslib was needed since it's a TUF dependency, and it's also needed for the FixtureBuilder we want to add.


tedbow (approved):
This looks good

The new version of TUF still has the problems of the keys changing order but we can figure how to deal with in another issue. This PR does make any worse.

PR #227, opened 2021-06-14
Port TUFTestFixtureRollbackAttack to FixtureBuilder

phenaproxima:
Just what it says! Ports the existing fixture to use the new FixtureBuilder class.


tedbow (approved):
looks good our check confirm the fixtures are the same but I also double checked and made sure the old way was not generating them.

PR #230, opened 2021-06-14
Port TUFTestFixtureUnsupportedDelegation to FixtureBuilder

phenaproxima:


generate_fixtures.py
```diff
@@ -185,20 +186,6 @@ def __init__(self):
             ['root', 'snapshot', 'targets', 'timestamp'])
         self.write_and_publish_repository()
 
-class TUFTestFixtureUnsupportedDelegation(TUFTestFixtureSimple):
-    # Sets up a repo using `path_hash_prefixes` which is currently not supported.
```
tedbow, line 189: We don't have this comment in the new `__init__.py` file. This is the important part about why are creating this fixture.

tedbow (changes requested):
Just need to copy over the 1 comment otherwise looks good.

generate_fixtures.py
phenaproxima: Fixed!

tedbow (approved):
The comment was added so looks good!

The fixture generation test passed so I assume everything else will.

approved if tests are Gree

PR #234, opened 2021-06-14
change testVerifiedDelegatedDownload to use a dataprovider. 

tedbow:
I was going to start making the tests for the fixtures in #216 
I realized we were going to have do tests like  `testVerifiedDelegatedDownload()`  but with a different fixture set

This PR: changes `testVerifiedDelegatedDownload()` to use a dataprovider which will allow testing other fixtures. to 

It also allows the test to confirm which  json files downloaded only for a specific target file because for each test case only 1 file is downloaded.


phenaproxima (approved):
Makes sense to me.

PR #235, opened 2021-06-14
Add extra arguments to testDelegationErrors() 

tedbow:
this will make  testing https://github.com/php-tuf/php-tuf/pull/216 easier and more robust

Add extra arguments to testDelegationErrors() 

This allows testing other fixtures.

Also adding `$expectedClientVersions` allows the test to confirm which roles were evaluated without this confirmation we don't know if a file was not found  because the role was not evaluated or evaluated and  file was still not found.

### DIFF
--- a/tests/Client/UpdaterTest.php
+++ b/tests/Client/UpdaterTest.php
@@ -424,8 +424,8 @@ class UpdaterTest extends TestCase
             // See § 5.6.7.2.1 and 5.6.7.2.2.
             'delegation is after terminating delegation' => [
                 'TUFTestFixtureNestedDelegatedErrors',
-                // @todo file file name in https://github.com/php-tuf/php-tuf/pull/216 becauses it does match the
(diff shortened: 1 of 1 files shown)


tests/Client/UpdaterTest.php
```diff
@@ -345,27 +353,96 @@ public function providerDelegationErrors(): array
...
+                'level_2_after_terminating_unfindable.txt',
+                [
+                    'root' => 6,
+                    'timestamp' => 6,
+                    'snapshot' => 6,
+                    'targets' => 6,
+                    // The client does not update the 'unclaimed.json' file because
+                    // the target file does not match the 'paths' property for the role.
+                    // @todo Update version in https://github.com/php-tuf/php-tuf/pull/216
+                    'unclaimed' => 1,
```
tedbow, line 438: If we had already been assert the client versions after the download attempt it would have been more obvious that the path for `unclaimed` did not match this target

phenaproxima (approved):

PR #236, opened 2021-06-15
Rename assertClientRepoVersions() and expectedClientVersions for clarity

phenaproxima:
I think that assertClientFileVersions() and expectedFileVersions would be clearer names for these things. This is a straight-up search and replace.


tedbow (approved):
Looks good!

PR #237, opened 2021-06-15
Create Composer script to generate fixtures

tedbow:
Just to make easier to update python dependencies and create fixtures


phenaproxima (changes requested):
To make sure it works, can you also modify our CI configuration so that it invokes this command, rather than calling `pipenv` directly?

Also: I don't think we actually need to call `pipenv update`, because if I'm reading the Pipfile correctly, all of our dependencies are locked to specific versions anyway.

tedbow (commented):
Great idea!

phenaproxima:
😂 It was _your_ idea, @tedbow!

phenaproxima (approved):
Approved after manual test by @tedbow.

PR #238, opened 2021-06-15
Remove unused targets in fixtures

tedbow:
Before we had the `Fixturebuilder` class we were using fixture classes that inherited from other fixture classes.

This meant that in addition to the specific targets that were added in a specific fixture class each class also had all the targets that were added in the inheritance chain. Because when we converted the fixtures to use `Fixturebuilder` we wanted to produce the exact same fixtures to ensure we were not losing/changing test coverage our fixture classes still have target that they don't use for tests.

For example `TUFTestFixtureNestedDelegatedErrors` has the targets added in `TUFTestFixtureNestedDelegated` and `TUFTestFixtureSimple` which use to be in inheritance chain of the classes.

For each fixture set we should remove all targets that aren't specifically tested with that fixture set


tedbow:
I changed this a draft PR because I only did `TUFTestFixtureNestedDelegatedErrors`  we need to make sure all fixtures are cleaned up

phenaproxima:
This has not been touched in years, and our fixtures just got a major overhaul in #366 (and I have even more ideas on that front). I don't think we need to keep this PR open.

PR #239, opened 2021-06-16
\Tuf\Metadata\Factory doesn't account for delegated roles

tedbow:
This adds test for the class in 1 commit and fixes the problem in the next

Discovered this in #216 because need to call `\Tuf\Metadata\TargetsMetadata::getRole()` and it was always returning 'targets' even for delegated roles.

### DIFF
--- a/tests/Metadata/FactoryTest.php
+++ b/tests/Metadata/FactoryTest.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace Tuf\Tests\Metadata;
+
+use Tuf\Metadata\Factory;
+use PHPUnit\Framework\TestCase;
+use Tuf\Metadata\RootMetadata;
+use Tuf\Metadata\SnapshotMetadata;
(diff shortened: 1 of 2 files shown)


tests/Metadata/FactoryTest.php
```diff
@@ -0,0 +1,56 @@
...
+     *   The role to load.
+     */
+    public function testLoad(string $class, string $role): void
+    {
+        $localRepo = $this->memoryStorageFromFixture('TUFTestFixtureDelegated', 'client/metadata/current');
+        $factory = new Factory($localRepo);
+
+        $metadata = $factory->load($role);
+        self::assertInstanceOf($class, $metadata);
+        self::assertEquals($class::TYPE, $metadata->getType());
```
phenaproxima, line 37: Can we use `assertSame()` instead?

phenaproxima (changes requested):

phenaproxima (approved):

PR #240, opened 2021-06-23
Idea: remove test fixtures from repo and ensure generation is always the same by using hash of fixtures directory

tedbow:
Just an idea...
(all PHP tests will fail currently because I need to save the fixtures for them again)
We currently have all the generated test fixtures back in the repo. In #225 we add test to make sure that running the `generate_fixtures.py` always produced the fixtures committed to the repo.

This is good because it means that we can guarantee that no change will happen to the fixtures without us knowing. This could happen if some update in the python TUF library makes some change we weren't aware of. It also helps if we change our `FixtureBuilder` class for something we **think** will not make a change in the produced fixtures but it actually does.

The problem is this creates huge PRs like #216 which is 1 reason we didn't want the fixtures in the repo in the first place.

This PR removes all the fixtures from the repo but creates a hash of the fixtures directory in `generate_fixtures.py` and saves to a fixtures-hash.txt file that is committed to the repo.

Because we have an existing github action to test that running `generate_fixtures.py does not produce any git changes this means that we will always catch changes if 
1. An update to Python TUF changes the fixtures
2. Someone updates `generate_fixtures.py` but does not run it and commit the change to the has file.

It also means any reviewer would made aware that a change is made in the fixtures.


fixtures/builder.py
```diff
@@ -40,6 +41,16 @@ def __init__(self, name):
 
         self.repository.status()
 
+    def __del__(self):
+        # Create a hash for the generated fixture.
+        hash_file = open(self.dir + "/hash.txt", "w")
+        file_contents = ""
+        for folder in ['client', 'server']:
+            file_contents = file_contents + folder + "-hash:" + dirhash(self.dir + "/" + folder, 'md5') + "\n"
```
phenaproxima, line 49: Why can't we just hash up the entire fixture directory, both `client` and `server`, in a single hash, before we create `hash.txt`?

fixtures/builder.py
```diff
@@ -40,6 +41,16 @@ def __init__(self, name):
...
+
+        n = hash_file.write(file_contents)
```
phenaproxima, line 51: `n` is not used.

generate_fixtures.py
```diff
@@ -25,4 +28,13 @@ def generate_fixtures():
     TUFTestFixtureThresholdTwoAttack.build()
 
 
+# Remove all previous fixtures.
+for f in glob.glob("fixtures/*/client"):
+    shutil.rmtree(f)
+for f in glob.glob("fixtures/*/server"):
+    shutil.rmtree(f)
+# Delete has files to ensure they are generated again.
```
phenaproxima, line 36:
```suggestion
# Delete hash files to ensure they are generated again.
```

generate_fixtures.py
```diff
@@ -25,4 +28,13 @@ def generate_fixtures():
...
+for f in glob.glob("fixtures/*/hash.txt"):
+    os.remove(f)
```
phenaproxima, line 38: IMHO it should be `FixtureBuilder`'s job to remove the existing hash file, maybe during `__init__`?

phenaproxima (changes requested):

fixtures/builder.py
tedbow: that makes sense. I think this was before I figured out we needed to delete all the hash files first.

I also moved the opening of the hash file till after the hash is create so we don't have the empty hash.txt file in the directory.

fixtures/builder.py
tedbow: fixed

generate_fixtures.py
tedbow: I thought about that but in case where we are removing a fixture or where developer locally built the fixtures on a different branch that had a different set of fixtures we still want all the files under `fixtures/` to be deleted regardless of whether they are in the current set of fixtures.

generate_fixtures.py
tedbow: fixed

fixtures/builder.py
```diff
@@ -40,6 +41,13 @@ def __init__(self, name):
 
         self.repository.status()
 
+    def __del__(self):
+        # Create a hash for the generated fixture.
+        file_contents = "hash:" + dirhash(self.dir, 'md5')
```
phenaproxima, line 46: Is the word `hash:` really necessary? Also, MD5 isn't supposed to be used any more, since it's vulnerable to rare collisions. Can we use SHA-256?

fixtures/builder.py
```diff
@@ -40,6 +41,13 @@ def __init__(self, name):
...
+        hash_file = open(self.dir + "/hash.txt", "w")
```
phenaproxima, line 47: We should use Python's `with open() as hash_file` construct here; if I remember correctly, it does automatic clean-up.

phenaproxima (changes requested):

fixtures/builder.py
tedbow: fixed

fixtures/builder.py
tedbow: I didn't know about `with open() as`
fixed

fixtures/builder.py
```diff
@@ -40,6 +41,12 @@ def __init__(self, name):
 
         self.repository.status()
 
+    def __del__(self):
+        # Create a hash for the generated fixture.
+        file_contents = dirhash(self.dir, 'sha256')
```
phenaproxima, line 46: Nit: `file_contents` is an unnecessary variable.

generate_fixtures.py
phenaproxima: Is this actually needed, though? Because in `FixtureBuilder.__del__`, we're overwriting the existing hash file.

phenaproxima (commented):
I really like it; just a nit and a question.

fixtures/builder.py
tedbow: Just to note. I thought that maybe we should use "md5" because it is faster. It is not collision resistant but I think is for deliberate attack. So someone could find a collision if the used a program to brute force and try to find a collision. The chance that we would produce 2 versions of a fixture folder that would *happen* to produce the same hash *and* those produced fixtures would *also* be valid json and pass tests would probably be impossible.

but I ran timer test on `generate_fixtures.py` I saw no consistent difference, probably because producing the hashes are a very small part of the total script time.

fixtures/builder.py
tedbow: If we get rid of the var and use
 
 `hash_file.write(dirhash(self.dir, 'sha256'))`
 
 This will actually change the hashes because it hashes the directory *after* we open the empty `hash.txt` file. I tested this to be sure.
 
 Changed the to `hash`. I think it is better to hash the directory with just the fixtures and not the empty `hash.txt` file.

phenaproxima: Oooh, interesting! Can you add a comment explaining this, so that we don't accidentally change it later?

tedbow:
@phenaproxima re

> Is this actually needed, though? Because in FixtureBuilder.__del__, we're overwriting the existing hash file.

See my comment about the hash.txt files themselves actually being hashed if they are present, even if empty. We delete them all first

fixtures/builder.py
tedbow: Added a comment

generate_fixtures.py
tedbow: See my comment about the hash.txt files themselves actually being hashed if they are present, even if empty.  We delete them all first because these would be the hashes for the fixtures before changes to builder scripts. Also there could be file for fixture we have deleted.

phenaproxima (approved):
Thanks! This is a nice change. Keeps our fixtures well-controlled, but without making the PRs giant. Great stuff.

fixtures/builder.py
```diff
@@ -40,6 +41,15 @@ def __init__(self, name):
 
         self.repository.status()
 
+    def __del__(self):
+        # Create a hash for the generated fixture. Generate the hash before
+        # we open the `hash.txt` file because the empty file will affect the
+        # hash of directory.
+        hashes = 'client:' + dirhash(self.dir + '/client', 'sha256')
+        hashes += "\nserver:" + dirhash(self.dir + '/server', 'sha256')
```
tedbow, line 49: I just realized that when we were just hashing `self.dir` that the hash would also be affected by the `__init__.py` files themselves, even if the changes didn't make any changes to the fixtures.  So if we add/updated comments or var names this would change the hash. So are reviewer to do a thorough job in these cases they would have to actually compare the generated fixtures to make sure they **hadn't actually changed**.

Also if we wanted to add README files to the fixtures this would change the hashes. 

So I think it is better only hash the `client` and `server` directories. I also putting them on separate lines. I think this will make review easier also. In the case where there are only expected changes to the server files this will prove that without the reviewer having to manually check the generated `client` folder.

phenaproxima: That's a good point. I agree with your reasoning and I think this change looks good.

PR #241, opened 2021-06-24
Add a Composer script to check test coverage

phenaproxima:


TravisCarden (approved):
That looks right to me!

PR #242, opened 2021-07-02
Add test_targets.json to each test fixture that has the expected outcome for attempt to download each target

tedbow:
Right we have to test providers for downloading targets `providerVerifiedDelegatedDownload()` and `providerDelegationErrors()`. This makes it complicated to all the test cases for 1 fixture. I had [1 idea](url) to simplify but it also may be simpler if the test cases for each fixture where in file with the fixture itself. 

I can see how this would work be test cases for attempt to download targets but haven't thought about it with other tests yet.

### DIFF
--- a/fixtures/TUFTestFixtureNestedDelegated/test_targets.json
+++ b/fixtures/TUFTestFixtureNestedDelegated/test_targets.json
@@ -1,0 +1,72 @@
+[
+  {
+    "target_path": "level_1_target.txt",
+    "find": true,
+    "expected_versions": {
+      "root": 5,
+      "timestamp": 5,
+      "snapshot": 5,
+      "targets": 5,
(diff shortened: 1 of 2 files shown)


phenaproxima:
Here's another possibility: we could split the test cases per-fixture into private methods of the test, and then have the data provider glue them together. So something like:

```
function provider(): array {
  return array_merge(
    $this->delegatedFixtures(),
    $this->moreCasesFromADifferentFixture(),
    $this->evenMoreCasesFromYetAnotherFixture()
}
```

I guess I'm not super certain if splitting this out into its own file confers much advantage. I'm open to being convinced, though.

phenaproxima:
The more I think about it, the more I like this idea. It brings us closer to having a generic conformance suite.

PR #243, opened 2021-07-07
Remove non_repo_fixtures directory

phenaproxima:
The stuff in here is only used to unit test the JsonNormalizer, and there's no real reason for them to be in their own files. Let's move the JSON into the test itself.


tedbow:
I actually like having the actual files but I think the folder name `non_repo_fixtures directory` is weird. Could we just move them to `/fixtures/json` since `/fixtures` is no longer in our `.gitignore` file?

tedbow (approved):
Looks good!

PR #244, opened 2021-07-07
Remove constants.php

phenaproxima:
Do we really need this file? I don't understand the usefulness. So, if everyone agrees, maybe we just remove it for now...


tedbow (approved):
Yeah I agree we aren't going to support other types any time soon.

If somebody else want to support other type will have to extend our classes and so they can override `getSupportedKeyTypes()`

PR #245, opened 2021-07-07
Rename PotentialAttackException namespace and base class

phenaproxima:
I think that the `Tuf\Exception\PotentialAttackException\PotentialAttackException` class name is needlessly verbose. The reason I think that is because, as far as TUF need concern itself, there's no difference between a "potential" attack and a _real_ attack -- there's no need to soft-pedal it. Therefore, I propose we consider all "potential" attacks to be regular attacks, and rename this to `Tuf\Exception\Attack\AttackException`.


src/Client/Updater.php
```diff
@@ -152,30 +152,30 @@ private static function getFileNameType(string $fileName): string
...
      *   Thrown if the signature threshold has not be reached.
+     *@todo The Python implementation has an optional flag to "unsafely update
+     *     root if necessary". Do we need it?
+     *     https://github.com/php-tuf/php-tuf/issues/21
+     *
+     * @see https://github.com/php-tuf/php-tuf/issues/21
+     *
+     * @todo Add docs. See python comments:
+     *     https://github.com/theupdateframework/tuf/blob/1cf085a360aaad739e1cc62fa19a2ece270bb693/tuf/client/updater.py#L999
+     *     https://github.com/php-tuf/php-tuf/issues/162
```
tedbow, line 178: Moving this comment is unrelated

tedbow (changes requested):
Looks good except for an unrelated change

tedbow (approved):
Looks good!

PR #250, opened 2021-07-08
Remove TestUpdater

phenaproxima:
We don't really need this class. Why not simply make the clock injectable into `Updater` directly?


tedbow (approved):
Looks good!

Just having the 2 files `UpdaterTest.php` and `TestUpdater.php` was a bit confusing.

PR #251, opened 2021-07-08
Move some overhead into getSystemInTest()

phenaproxima:
This PR does three interrelated things:

1. It makes `getSystemInTest()` set up `$this->testRepo`.
2. And `$this->localRepo`.
3. It makes `getSystemInTest()` always assert the hard-coded starting versions, for extra safety and consistency.

### DIFF
--- a/tests/Client/UpdaterTest.php
+++ b/tests/Client/UpdaterTest.php
@@ -127,6 +127,8 @@ class UpdaterTest extends TestCase
             ],
         ];
 
+        $this->testRepo = new TestRepo($fixturesSet);
+
         // Remove all '*.[TYPE].json' because they are needed for the tests.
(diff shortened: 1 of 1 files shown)


tedbow:
This is great!

Should we add extra comments in `getSystemInTest()` to be clear it is going to set up class properties? So it is clear what effect it has. 

Other option would be that instead of returning `$updater` the method could set up `$this->localRepo`, `$this->testRepo` *and* `$this->updater` and then change the name to something like `setUpdateTestCaseForFixture()`

phenaproxima:
I'd prefer not to add additional mutable state to the class, but I agree that a comment is handy. So, added that.

tedbow (approved):
Great simplification!

PR #252, opened 2021-07-08
Create a FixturesTrait with all necessary methods to interact with fixtures

phenaproxima:
I think it would make sense to have a trait that contains all methods that tests need to interact with and assert the validity of our test fixtures. This could replace parts of UtilsTrait, and totally replace MemoryStorageLoaderTrait.


tests/TestHelpers/FixturesTrait.php
```diff
@@ -0,0 +1,208 @@
...
+     * @param ?int[] $expectedVersions
+     *   The expected versions. The keys are the file names, without the .json
+     *   suffix, and the values are the expected version numbers, or NULL if
+     *   the file should not be present.
+     * @param \ArrayAccess $storage
+     *   The durable storage for the metadata.
+     *
+     * @return void
+     */
+    protected function assertMetadataVersions(array $expectedVersions, \ArrayAccess $storage): void
```
tedbow, line 176: could be static

tests/TestHelpers/FixturesTrait.php
```diff
@@ -0,0 +1,208 @@
...
+     * @param string $fixtureName
+     *     The name of the fixture to use.
+     * @param string $path
+     *     The relative path (within the fixtures directory) for the client
+     *     data.
+     *
+     * @return MemoryStorage
+     *     Memory storage containing the test client data.
+     */
+    private function loadFixtureIntoMemory(string $fixtureName, string $path = 'client/metadata/current'): MemoryStorage
```
tedbow, line 118: could be static

tedbow (changes requested):
I like the trait. a couple nits

tests/TestHelpers/FixturesTrait.php
phenaproxima: Fixed.

tests/TestHelpers/FixturesTrait.php
phenaproxima: Fixed.

tedbow (approved):
👍

PR #254, opened 2021-07-09
Updater always thinks that keys have been rotated

phenaproxima:


tests/Client/UpdaterTest.php
```diff
@@ -1084,15 +1084,18 @@ public function providerRefreshException(): array
...
+        // not trying to complete the refresh.
         $this->serverStorage->removeRepoFile($fileName);
         try {
             $updater->refresh();
+            $this->fail('No RepoFileNotFound exception thrown');
         } catch (RepoFileNotFound $exception) {
-            $this->assertSame("File $fileName not found.", $exception->getMessage());
-            $this->assertClientFileVersions($expectedUpdatedVersions);
-            return;
+            // We don't have to do anything with this exception; we just wanted
```
tedbow, line 1095: Maybe mention the exception message comes from `TestRepo` so there is not point in asserting a message produced by the test class.

tests/Client/UpdaterTest.php
```diff
@@ -1138,6 +1141,10 @@ public function providerFileNotFoundExceptions(): array
             ],
             [
                 'TUFTestFixtureSimple',
+                // Deleting timestamp.json and 1.snapshot.json from the server will cause Updater::updateTimestamp()
+                // and Updater::refresh() to error out. That's fine in these cases, because we're not trying to finish
+                // the refresh. This will implicitly check that Updater::updateRoot() doesn't erroneously think that
+                // keys have been rotated, and therefore delete the local timestamp.json and snapshot.json.
```
tedbow, line 1147: Since we have separate test method for this now maybe we could have very short with a @see to the new test method?

tests/Client/UpdaterTest.php
```diff
@@ -1317,4 +1324,62 @@ public function providerAttackRepoException(): array
...
+                    'timestamp' => null,
+                    'snapshot' => null,
+                    'targets' => 1,
+                ],
+            ],
+        ];
+    }
+
+    /**
+     * Tests that the updater correctly handles key rotation.
```
tedbow, line 1362: either in the test provider cases or the comment here we should mention that we expect the timestamp and snapshot files to be delete if *either* keys are rotated.

Also we should reference the part in the spec that we are covering as is now in main after #248
Also we could add `@covers`'s  for `\Tuf\Client\Updater::updateRoot()` and `\Tuf\Client\Updater::hasRotatedKeys()`

tedbow (changes requested):
mainly asking for doc changes

tests/Client/UpdaterTest.php
phenaproxima: Done.

tests/Client/UpdaterTest.php
phenaproxima: I kept it as-is, but added the `@see`. I figure this is something that is complex enough to justify explaining it verbosely.

tests/Client/UpdaterTest.php
phenaproxima: Done!

PR #257, opened 2022-03-20
Make fields > delegations optional according to specifications

MagnusSinger:
According to the TUF specifications, the fields > delegations field on the targets file should be optional (https://theupdateframework.github.io/specification/latest/#file-formats-targets). This can be accomplished by changing the getSignedCollectionOptions method and making $options['fields']['delegations'] an Optional instead of Required


phenaproxima:
This was fixed by #302 😎

PR #258, opened 2022-03-20
Respect metadata file lengths defined in snapshot and timestamp metadata

MagnusSinger:
According to the specifications, the length and hashes fields should be optional and not forbidden (https://theupdateframework.github.io/specification/latest/#file-formats-snapshot). We worked with the tuf-go server package instead of the tuf-python package, and this approach breaks this workflow. Because it's not intended by the specifications either, I would not disallow these fields.


phenaproxima:
This will need automated test coverage, but otherwise it looks like TargetsVerifier is implementing the proper behavior according to the spec. (Always verify the versions, but verify the hashes if available.)

That said, one missing piece here is that, as we fetch and verify targets metadata, we are using Updater::MAXIMUM_DOWNLOAD_BYTES to constrain the download sizes, rather than the length in the snapshots metadata, if it's defined. We should probably fix that in a separate PR.

phenaproxima:
After doing some experimenting, I'll work directly on this PR, rather than in #286.

phenaproxima:
Changing the scope here a bit.

Let's leave hashes unsupported for now, but add support for the `length` field.

tests/Client/TestRepo.php
```diff
@@ -25,6 +25,8 @@ class TestRepo implements RepoFileFetcherInterface
      */
     public $fileContents = [];
 
+    public array $fetchMetadataArguments = [];
```
tedbow, line 28: This really not clear what this is for. 

I am a kinda of understand because I looking at the specific PR where you are adidng. but if I came back in a year and was looking at the codebase in another context and stumble upon this  I think I would throughly confused.

tedbow: so really for this and anywhere it is used but generally, the declaration, the setting and reading have a relationship and reasoning but it isn't explained

tedbow (changes requested):
for comments/docs requested in my comment

tests/Client/TestRepo.php
phenaproxima: Agreed. I added a comment.

tests/Client/UpdaterTest.php
```diff
@@ -1325,4 +1325,47 @@ public function testKeyRotation(string $fixtureName, array $expectedUpdatedVersi
...
+        $this->getSystemInTest($fixtureName)->refresh();
+
+        $fetchMetadataArguments = [];
+        foreach ($this->serverStorage->fetchMetadataArguments as [$fileName, $maxBytes]) {
+            $fetchMetadataArguments[$fileName] = $maxBytes;
+        }
+        // The length of the timestamp metadata is never known in advance, so it
+        // is always downloaded with the maximum length.
+        $this->assertSame(Updater::MAXIMUM_DOWNLOAD_BYTES, $fetchMetadataArguments['timestamp.json']);
+        $this->assertSame($expectedLength, $fetchMetadataArguments[$downloadedFileName]);
```
tedbow, line 1369: @phenaproxima do you think we need a test proving that if the length specified didn't match the file length that was attempted to be download it cause an error. 

What is the purpose of the optional field ? what is protecting against?

phenaproxima: I think we should, but that should be part of the GuzzleFileFetcher test (if we don't have something already), since that is where the limit is enforced. I'm thinking that should belong in a separate PR.

The purpose of the length field is, as far as I know, to protect against infinite-data DoS attacks.

phenaproxima: Never mind - \Tuf\Tests\Client\UpdaterTest::testVerifiedDownload() already has a test like that:

```
        // If the stream is longer than expected, we should get an exception,
        // whether or not the stream's length is known.
        $stream = $stream = $this->prophesize('\Psr\Http\Message\StreamInterface');
        $stream->getSize()->willReturn(1024);
        $this->serverStorage->fileContents['testtarget.txt'] = new FulfilledPromise($stream->reveal());
        try {
            $updater->download('testtarget.txt')->wait();
            $this->fail('Expected DownloadSizeException to be thrown, but it was not.');
        } catch (DownloadSizeException $e) {
            $this->assertSame("testtarget.txt exceeded 24 bytes", $e->getMessage());
        }
```

phenaproxima: That being said, we should _also_ have a test against actual infinite data attacks as enforced by Guzzle. That, I think, should be in a separate PR.

tedbow: So we would have unit tests for `GuzzleFileFetcher` but we won't have integration test that if you length was set to something smaller than the actual file you in fact would not get the whole file?

tedbow: hmm. let look at this more

phenaproxima: Looking at this closer, I see (and remember) why we're doing things the way we're doing.

In GuzzleFileFetcher, we implement a callback that is called every so often and verifies the maximum length as it comes through the network. However...**that only works with cURL**, which is why we can't really unit test it.

Therefore, the updater also do a length check _after_ the transfer is finished. That, we _do_ test (as I pointed out previously).

The weakness here, I think, is that if we try to download the entire response before processing it, we might still be vulnerable to infinite data attacks. What I think might be a solution to that is to take advantage of https://docs.guzzlephp.org/en/stable/request-options.html#stream, which (if I understand it correctly) means that Guzzle won't try to wait for the entire response to come in before fulfilling the promise. That would allow TUF to then read the maximum number of bytes from the stream, before it's "finished", and verify that the expected size is correct.

```
we won't have integration test that if you length was set to something smaller than the actual file you in fact would not get the whole file?
```

We don't have a test like that. I don't think we care if the transferred file is smaller; we only care if it's too big. That said, I think it would be wise for us to harden that and check that the transferred size is actually as big as we expected (assuming the length was known). Definitely not in scope here.

tedbow: @phenaproxima ok I think what I was expecting was coverage like we have in `\Tuf\Tests\Client\UpdaterTest::testVerifiedDownload` because in the new case and in this case I think `\Tuf\Client\Updater::checkLength` could throw a `DownloadSizeException` but we aren't testing that case here. and since that is not part of `Updater` it should be test in `UpdaterTest` I think

tedbow (changes requested):
for test coverage

tedbow (approved):
Looks good!

PR #259, opened 2022-03-20
Allow spec version "1.0" for compatibility with older TUF implementations

HLeithner:
Based on the specification the spec_version has to be x.y.z sadly some implementations generates x.y entries.

More information can be found in https://github.com/theupdateframework/go-tuf/issues/206#issuecomment-1020213003

### DIFF
--- a/src/Metadata/MetadataBase.php
+++ b/src/Metadata/MetadataBase.php
@@ -3,10 +3,12 @@
 namespace Tuf\Metadata;
 
 use Symfony\Component\Validator\Constraints\All;
+use Symfony\Component\Validator\Constraints\AtLeastOneOf;
 use Symfony\Component\Validator\Constraints\Collection;
 use Symfony\Component\Validator\Constraints\Count;
 use Symfony\Component\Validator\Constraints\DateTime;
 use Symfony\Component\Validator\Constraints\EqualTo;
+use Symfony\Component\Validator\Constraints\IdenticalTo;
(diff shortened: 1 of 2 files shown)


phenaproxima:
That sounds like a bug in go-tuf...? Our goal is generally to obey the spec, so if there's no reason go-tuf _can't_ fix this problem, I'm not sure if we should change this.

HLeithner:
actually it's not go-tuf it's primary rust-tuf...

SniperSister:
@phenaproxima looking at other implementations (https://github.com/theupdateframework/rust-tuf/blob/d0c3ef9d7ac512fed069d97059d1160eaeac0efe/tuf/src/pouf/pouf1/shims.rs#L20) there is a very valid point for relaxing the validation being made:

> We also need to handle the literal "1.0" here, despite that fact that it is not a valid version according to the SemVer spec, **because it is already baked into some of the old roots.**

So even IF tuf-go and tuf-rust would update their implementations, existing metadata sets might be around for a long time. Relaxing the spec is the only option for php-tuf to consume these existing repositories.

SniperSister:
@phenaproxima any update and/or decision?

phenaproxima:
I have no problem with doing this.

Since the spec has not reached version 1.1.0 yet, I think it's fair either allow `1.0`, or a real semver version that begins with `1.0.`.

tedbow (approved):
🤖

PR #263, opened 2022-05-11
Add workflow responsible for notifying of new TUF spec releases

rdimitrov:
The following PR adds a workflow that keeps track and creates an issue for notifying of new TUF specification releases.

The version (v1.0.29) of the spec which is currently used for the verification is added without verifying if this is indeed the version that the `php-tuf` project is compliant with. In that sense, please advise which version to use and I'll update the PR accordingly.

**Note:** Dependent on - https://github.com/theupdateframework/specification/pull/224

Signed-off-by: Radoslav Dimitrov <dimitrovr@vmware.com>

### DIFF
--- a/.github/workflows/specification-version-check.yml
+++ b/.github/workflows/specification-version-check.yml
@@ -1,0 +1,14 @@
+on:
+  schedule:
+    - cron: "0 13 * * *"
+  workflow_dispatch:
+name: Specification version check
+jobs:
+  # Get the latest TUF specification release and open an issue (if needed)
+  specification-bump-check:
+    permissions:
+      contents: read
+      issues: write
+    uses: theupdateframework/specification/.github/workflows/check-latest-spec-version.yml@master
+    with:
+      tuf-version: "v1.0.29" # Should be updated to the version the project supports either manually or extracted automatically. You can see how python-tuf did that as an example.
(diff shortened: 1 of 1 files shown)


rdimitrov:
Hey, just so you know the dependent PR is now merged in the specification repo 👍

phenaproxima:
+1 for this idea; looping @tedbow and @davidstrauss in for review.

davidstrauss (approved):

.github/workflows/specification-version-check.yml
```diff
@@ -0,0 +1,14 @@
...
+name: Specification version check
+jobs:
+  # Get the latest TUF specification release and open an issue (if needed)
+  specification-bump-check:
+    permissions:
+      contents: read
+      issues: write
+    uses: theupdateframework/specification/.github/workflows/check-latest-spec-version.yml@master
+    with:
+      tuf-version: "v1.0.29" # Should be updated to the version the project supports either manually or extracted automatically. You can see how python-tuf did that as an example.
```
phenaproxima, line 14: I think we need to bring this down to the actual last version of the spec we specifically supported, which...I can't remember what it is. But we should change this before we merge.

rdimitrov: That makes sense 👍 Let me know which is it and I'll update the PR accordingly

phenaproxima: Sorry for the very long delay. Our current version of the spec is 1.0.18. If you can make this change in the PR, I will happily merge it.

rdimitrov: Thanks, @phenaproxima! Just did it 👍

PR #264, opened 2022-06-12
Make guzzlehttp client a composer suggestion

SniperSister:
The current TUF client implementation has a hardcoded dependency for the guzzlehttp client even though the FileFetcherInterface defines a generic (guzzle) promise as a return type and therefore isn't bound to guzzle as http client. This PR makes the http client a suggestion, allowing downstream projects to either use Guzzle/Http or their own implementation of a file fetcher without having the hardcoded dependency wired into this package.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,12 +15,13 @@
         "ext-json": "*",
         "paragonie/sodium_compat": "^1.13",
         "symfony/validator": "^4.4 || ^5",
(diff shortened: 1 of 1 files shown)


phenaproxima:
I think this makes sense. I need to make a few changes, though, in order to fix our CI jobs. Since I can't push to the fork, I need to close this in favor of #272.

PR #266, opened 2022-07-11
Collapse php-tuf/phpcodesniffer-standard into project

TravisCarden:
The only other project that used `php-tuf/phpcodesniffer-standard` has diverged, so there's no reason to have a separate library for it. This collapses it directly into this project--no changes to the standard itself. Once this is merged, we can delete the whole `php-tuf/phpcodesniffer-standard` repo.


phenaproxima (approved):
No objections here.

Issue #265, opened 2022-06-23
Support TUF repos that do not use consistent snapshots


ergonlogic:
Related to #98.

We currently enforce consistent snapshots in two ways, by:
1. validating that root metadata has consistent snapshots enabled (in [src/Metadata/RootMetadata.php](https://github.com/php-tuf/php-tuf/blob/main/src/Metadata/RootMetadata.php#L44)), and
2. throwing exceptions when consistent snapshots is not enabled (in [src/Client/Updater.php](https://github.com/php-tuf/php-tuf/blob/main/src/Client/Updater.php))

I've started on a fix in my [fork over on gitlab](https://gitlab.com/drupal-infrastructure/package-signing/php-tuf-client/-/commit/cb1056e20e8a8dcd8cabb9a96fc28c7334e65f4a).

phenaproxima:
This was done in #278.

ergonlogic:
In ergonlogic/php-tuf@fc2608a7, I added a change ("Make delegations optional.") that wasn't strictly related to this issue. I'll open a new issue to address that.

PR #268, opened 2022-12-14
Relates to #265: Make consistent snapshots optional

ergonlogic:


ergonlogic:
sorry, I will re-apply the change at the tip of main, rather than whatever's happened here.

PR #269, opened 2022-12-14
Relates to #265: Do not require consistent snapshots.

ergonlogic:


ergonlogic:
This appears to force 'consistent snapshot' to be true. When we test this with `"consistent_snapshot": false,` in `1.root.json`, we get:

```
> pre-file-download: Tuf\ComposerIntegration\Plugin->preFileDownload
                                                                                       
  [Tuf\Exception\MetadataException]                                                    
  Object(ArrayObject)[signed][consistent_snapshot]:                                    
      This value should be equal to true. (code 478618a7-95ba-473d-9101-cabd45e49115)
```

src/Client/Updater.php
```diff
@@ -103,7 +103,13 @@ class Updater
      */
     protected $universalVerifier;
 
-    /**
+  /**
+   * The root metadata.
+   * @var \Tuf\Metadata\MetadataBase|null
+   */
+    private $rootMetadata;
```
phenaproxima, line 110: Do we really need to keep the root metadata as a class property? I feel like this is a way to accidentally introduce side effects later.

phenaproxima (changes requested):
I think that there are generally two things missing here: a generous sprinkling of comments, and test coverage.

phenaproxima:
Superseded by #278.

PR #270, opened 2023-01-10
Loosen Python dependency constraints

phenaproxima:
I was unable to get `pipenv install` working on my M1 Pro laptop, under macOS Ventura. (The errors I was getting were too long, and too vague, to reproduce here.) @tedbow encountered the same problem, and sent me his `Pipfile` and `Pipfile.lock`, which solved the problem by...basically removing all dependency version constraints.

That would be dangerous on its face except for the fact that these files are used only for testing. So, in the interest of unblocking development...


Pipfile
```diff
@@ -6,23 +6,23 @@ verify_ssl = true
...
-urllib3 = "==1.25.10"
-PyNaCl = "==1.4.0"
+urllib3 = "*"
+PyNaCl = "*"
 flake8 = "*"
 dirhash = "*"
 
 [requires]
-python_version = "3.9"
+python_version = "3.8"
```
davidstrauss, line 28: This change is not simply a weakening but a re-pinning on an older Python release. We could consider dropping this entirely if it needs to be more flexible, as `Pipfile` does not allow a more flexible (but specified) constraint. Dependencies will impose implicit constraints, anyway. The Pipenv developers have [defended this behavior as a best practice](https://github.com/pypa/pipenv/issues/2683).

Pipfile
```diff
@@ -6,23 +6,23 @@ verify_ssl = true
...
+pycparser = "*"
+python-dateutil = "*"
+requests = "*"
+securesystemslib = "*"
+six = "*"
 tuf = "==0.19.0"
-urllib3 = "==1.25.10"
-PyNaCl = "==1.4.0"
+urllib3 = "*"
+PyNaCl = "*"
```
davidstrauss, line 23: I would prefer we weaken the dependency specifications by providing floors or other bounded constraints on versions rather than unlocking the constraints entirely. If we need a lower minimum on some dependencies than today's, I understand, but it's risky for test reproducibility to unlock them entirely.

How about we bound them to versions that include (1) the explicit ones in `Pipfile` today and (2) the versions you found installed when you unlocked the constraints on macOS? That should fix your workflow today and provide an iterative path for resolving future issues without opening the barn doors fully.

The syntax for defining constraints in this file is specified in [PEP 440](https://peps.python.org/pep-0440/).

davidstrauss (changes requested):
I've provided detailed feedback in the diff. In short, I'm wary of unlocking the constraints entirely. I'd also prefer to not pin to Python 3.8 when we're already pinned to 3.9 for GitHub Actions test fixture generation.

Pipfile
phenaproxima: @davidstrauss I've looked through that PEP and damned if I can find how to do an "or" type of constraint.

Pipfile
phenaproxima: Reverted.

Pipfile
phenaproxima: I went ahead and did that. If you think that's acceptable, I think that would probably unblock things enough for now.

davidstrauss (approved):

PR #271, opened 2023-01-12
Drop support for PHP 7

phenaproxima:
I propose we drop support for PHP 7, for the following reasons:

* It is EOL. Since this is a security-related, cryptographically-sensitive project, it is a security risk for us to support dead versions of PHP.
* If someone uses PHP-TUF with `ext-sodium` installed, we don't want to inadvertently be using an unsupported, potentially insecure version of the Sodium library built into PHP.
* From the Drupal perspective, we don't really need to require PHP-TUF for _attended_ updates. Therefore, people could still use the Automatic Updates module with PHP 7.4; they'd just need to do their updates in the UI.
* Our tests use Prophecy, which requires PHP 7.3 or later. Additionally, we're currently supporting two major versions of PHPUnit, which leads to awkwardness around deprecated assertions. (This means we either need to drop PHP 7.2 support only, or alter our tests to use PHPUnit mocks instead of Prophecy. Until we do, our CI jobs won't pass.)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,8 @@
         "phpunit/phpunit": "^9",
         "symfony/phpunit-bridge": "^5",
         "squizlabs/php_codesniffer": "^3.7",
-        "slevomat/coding-standard": "^8.2"
+        "slevomat/coding-standard": "^8.2",
+        "phpspec/prophecy-phpunit": "^2"
     },
     "license": "MIT",
     "minimum-stability": "dev",
@@ -18,8 +19,7 @@
         "symfony/validator": "^4.4 || ^5",
         "guzzlehttp/guzzle": "^6.5 || ^7.2",
         "myclabs/deep-copy": "^1.10.2",
-        "guzzlehttp/psr7": "^1.7",
-        "phpspec/prophecy-phpunit": "^2"
+        "guzzlehttp/psr7": "^1.7"
     },
     "suggest": {
         "ext-sodium": "Provides faster verification of updates"
--- a/composer.json
+++ b/composer.json
@@ -3,7 +3,7 @@
     "description": "PHP implementation of The Update Framework (TUF)",
     "type": "library",
     "require-dev": {
-        "phpunit/phpunit": "^8.5.8|^9",
+        "phpunit/phpunit": "^9",
         "symfony/phpunit-bridge": "^5",
         "squizlabs/php_codesniffer": "^3.7",
         "slevomat/coding-standard": "^8.2"
@@ -12,7 +12,7 @@
     "minimum-stability": "dev",
     "prefer-stable": true,
     "require": {
-        "php": ">=7.2.5",
+        "php": "^8",
         "ext-json": "*",
         "paragonie/sodium_compat": "^1.13",
         "symfony/validator": "^4.4 || ^5",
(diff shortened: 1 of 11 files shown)


composer.json
```diff
@@ -3,16 +3,17 @@
     "description": "PHP implementation of The Update Framework (TUF)",
     "type": "library",
     "require-dev": {
-        "phpunit/phpunit": "^8.5.8|^9",
+        "phpunit/phpunit": "^9",
```
phenaproxima, line 6: I'm assuming the only reason we'd need to support PHPUnit 8 is for PHP 7.2 support.

composer.json
```diff
@@ -21,7 +22,7 @@
         "guzzlehttp/psr7": "^1.7"
     },
     "suggest": {
-        "ext-libsodium": "Provides faster verification of updates"
+        "ext-sodium": "Provides faster verification of updates"
```
phenaproxima, line 25: `ext-libsodium` was _never_ correct, as it turns out. `composer show --platform` shows that the correct extension name is `ext-sodium`, and the _library_ name is `lib-libsodium`. (Not sure what the difference is, exactly, but there is definitely no `ext-libsodium`.)

src/Client/DurableStorage/DurableStorageAccessValidator.php
```diff
@@ -55,7 +55,7 @@ protected function throwIfInvalidOffset($offset): void
     /**
      * {@inheritdoc}
      */
-    public function offsetExists($offset)
+    public function offsetExists($offset): bool
```
phenaproxima, line 58: This brings things in line with return types added to ArrayAccess in PHP core.

src/KeyDB.php
```diff
@@ -91,7 +91,7 @@ public function addKey(string $keyId, Key $key): void
         if (! in_array($key->getType(), self::getSupportedKeyTypes(), true)) {
             // @todo Convert this to a log line as per Python.
             // https://github.com/php-tuf/php-tuf/issues/160
-            throw new InvalidKeyException("Root metadata file contains an unsupported key type: \"${keyMeta['keytype']}\"");
+            throw new InvalidKeyException("Root metadata file contains an unsupported key type: \"{$keyMeta['keytype']}\"");
```
phenaproxima, line 94: The previous way threw deprecation errors on PHP 8.2.

phenaproxima (commented):
A little self-review.

tedbow:
This seems good to me. I can review it fully when the tests are passing

phenaproxima:
@tedbow, the current failures have to do with older versions of dependencies that are not compatible with PHP 8, and should therefore either be bumped, or dropped. IMHO that belongs in another PR; do you agree?

phenaproxima:
To be clear, here's the path I see to fully passing tests.

* After this is merged, attack #264, since the Guzzle client is the source of most of the current test failures. We should only be _actually_ depending on Guzzle's promises library, since that's the only part PHP-TUF truly needs in order to function.
* Move the Guzzle client to our dev dependencies (same constraint as we already have), and only testing Guzzle 6 in the "build lowest version" CI job. That would ensure we're still marginally supporting Guzzle 6, but not testing it on a version of PHP we _know_ will cause failures. (We'll have to revisit this once PHP 8.1 is the lowest supported version.)

tedbow (approved):
Ok looks good with the follow-up!

PR #272, opened 2023-01-12
Make the Guzzle HTTP client a Composer suggestion and move it to dev dependencies

phenaproxima:
Supersedes #264.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,7 +77,6 @@ jobs:
         dependencies:
           - 'symfony/validator:^4.4'
           - 'symfony/validator:^5'
-          - 'guzzlehttp/guzzle:^7.2'
         experimental: [false]
         include:
           - operating-system: ubuntu-latest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,8 +75,8 @@ jobs:
         operating-system: [ubuntu-latest, macos-latest, windows-latest]
         php: ['8.0', '8.1', '8.2']
         dependencies:
-          - 'symfony/validator:^4.4'
-          - 'symfony/validator:^5'
+          - '"symfony/validator:^4.4"'
+          - '"symfony/validator:^5"'
         experimental: [false]
         include:
           - operating-system: ubuntu-latest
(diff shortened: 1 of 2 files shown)


phenaproxima:
To summarize these changes:

* This replaces the runtime dependency on guzzlehttp/guzzle with the runtime dependencies we're truly using: guzzle/promises and psr/http-message.
* We still suggest people install guzzlehttp/guzzle, since that means they can use our GuzzleFileFetcher.
* Since we test that fetcher, we have guzzlehttp/guzzle as a dev dependency, along with guzzlehttp/psr7 (which is used in our test code).
* In nearly all CI jobs, we allow Composer to install Guzzle 7. However, the "build lowest version" will use Guzzle 6, because it runs Composer with the `--prefer-lowest` flag.

.github/workflows/build.yml
```diff
@@ -74,11 +74,11 @@ jobs:
...
         php: ['8.0', '8.1', '8.2']
+        # We need to use the ~ operator to prevent weirdness on Windows.
+        # See https://github.com/composer/composer/issues/10943
         dependencies:
-          - 'symfony/validator:^4.4'
-          - 'symfony/validator:^5'
-          - 'guzzlehttp/guzzle:^6.5'
-          - 'guzzlehttp/guzzle:^7.2'
+          - 'symfony/validator:~4.4'
+          - 'symfony/validator:~5.0'
```
tedbow, line 81: since the commit message for this was "specify 5.0 with tilde to prevent updating to symfony 6" but drupal 10 I think requires symfony 6 would this prevent php-tuf from being used on drupal 19?

phenaproxima: The other thing that prevents Symfony 6 (at the moment) is the requirement on `symfony/validator:^4.4 || ^5` in composer.json. That needs to be fixed in a follow-up PR, because you're right: until we allow Symfony 6, PHP-TUF is incompatible with Drupal 10.

tedbow: I am don't know enough about the interdepedencies of `symfony/*` packages but this jump out at me

tedbow (commented):
Looks good!

tedbow (approved):
🎉

SniperSister:
Perfect, that helps a lot :)

PR #274, opened 2023-01-12
Compress generated fixtures to speed up builds and prevent upload/download failures

phenaproxima:
I've been seeing some distressing behaviors happening during our CI runs -- uploads and downloads of the generated fixtures will sometimes hang, which requires me to go in and manually cancel and re-run the jobs.

I suspect this may be happening because, according to https://github.com/marketplace/actions/upload-a-build-artifact#too-many-uploads-resulting-in-429-responses, the fixtures are being uploaded as a directory, rather than an archive. With over 1,000 files in the fixtures, it's no wonder that this is causing trouble.

To solve this, I propose that the fixtures be packaged as an archive and uploaded/downloaded that way.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -165,7 +165,10 @@ jobs:
         uses: actions/download-artifact@v2
         with:
           name: fixtures
-          path: fixtures
+          path: fixtures.tar.gz
+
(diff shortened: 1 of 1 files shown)


tedbow (approved):
Looks good if all the tests pass

PR #275, opened 2023-01-13
Run CI workflow every day

phenaproxima:
We should be running our tests every day automatically. Not really controversial. This adds syntax to do that at 3 AM UTC, according to the documentation at https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule.


tedbow (approved):
Looks good

PR #277, opened 2023-01-18
Create consistent and inconsistent variants of all fixtures

phenaproxima:
Before we can merge #269, we're going to want to have versions of our fixtures that use consistent snapshots, and in(?)consistent snapshots.

This PR introduces a new Python class, `ConsistencyVariantFixtureBuilder`, which is essentially a wrapper around two FixtureBuilders that forwards every call to FixtureBuilder's public methods into a consistent fixture, and an inconsistent fixture.

The consistent fixtures generated from this are still identical to what they previously were -- you'll see that every `fixtures/*/consistent/hash.txt` is merely a rename of `fixtures/*/hash.txt`, with no changes. (The inconsistent fixtures all have new `hash.txt` files, which is expected, since the generated data is actually different.)

This also changes the PHPUnit tests to only look at the consistent fixtures for now. (We'll need to change that as part of #269.)


tedbow (approved):
Looks good

PR #278, opened 2023-01-19
Don't require consistent snapshots

phenaproxima:
#269 seems like there are unrelated changes in it, so to keep things as tightly scoped as possible, in this PR I'm going to support non-consistent snapshots, and ensure all the non-consistent fixtures are passing in our existing test cases.


phenaproxima:
This one is tricky to get right. The goal of ensuring all our existing test coverage also applies to the non-consistent fixtures mean that I had to try and refactor the UpdaterTest into two overall variants: one that looks only at the consistent fixtures, and one that looks at the non-consistent ones.

I decided ultimately to keep using data providers in UpdaterTest, but with deterministic and  helpful (I hope) keys that can be used to override the data sets in subclasses when needed. And it _is_ needed, since non-consistent snapshots will often have different version numbers for their metadata than the consistent snapshots.

As part of this, I also discovered PHPUnit's `@testdox` annotation, which helps make a lot clearer what is being tested, even when data providers are involved. I think it makes sense to always use this format, honestly, so I've made those changes (along with some other changes in the tests to improve that form of documentation).

tests/Client/UpdaterTest.php
```diff
@@ -536,7 +524,7 @@ public function providerVerifiedDelegatedDownload(): array
      *
      * § 5.6.7.1
      */
-    public function testMaximumRoles(): void
+    public function testRoleDownloadsAreLimited(): void
```
phenaproxima, line 527: When I renamed test methods like this, it was to make them clearer in testdox output (which also tends to make them clearer in code as well).

tests/Client/UpdaterTest.php
```diff
@@ -825,9 +801,9 @@ public function providerDelegationErrors(): array
      * @param array $expectedUpdatedVersions
      *   The expected updated versions.
      *
-     * @return void
-     *
      * @dataProvider providerRefreshRepository
+     *
+     * @testdox Refresh $fixtureName repository
```
phenaproxima, line 806: In testdox, this will read like `Refresh Delegated repository`. Very clear!

tests/Client/UpdaterTest.php
```diff
@@ -938,9 +912,9 @@ protected function assertClientFileVersions(array $expectedVersions): void
      * @param array $expectedUpdatedVersions
      *   The expected repo file version after refresh attempt.
      *
-     * @return void
-     *
      * @dataProvider providerExceptionForInvalidMetadata
+     *
+     * @testdox Invalid metadata in $fileToChange raises an exception
```
phenaproxima, line 917: We could make this even clearer (something like `Setting signed:version to 6 in root.json raises an exception`), but that is best done in a follow-up, since it would require me to refactor the test logic slightly.

tests/Client/UpdaterTest.php
```diff
@@ -979,28 +953,14 @@ public function providerExceptionForInvalidMetadata(): array
                     'targets' => 2,
                 ],
             ],
-            [
-                // § 5.3.4
-                '4.root.json',
```
phenaproxima, line 984: This test case doesn't apply to the inconsistent fixture, since it doesn't have a `4.root.json`, so it only appears in `ConsistentFixturesUpdaterTest`.

tests/Client/UpdaterTest.php
```diff
@@ -1243,12 +1196,27 @@ public function testUpdateRefresh(): void
         $updater->refresh(true);
     }
 
+    public function providerUnsupportedRepo(): array
```
phenaproxima, line 1199: Even though there's only one test case here, it needs to be moved to a data provider so that the subclasses can override the details as needed.

tests/Client/UpdaterTest.php
```diff
@@ -1280,65 +1240,36 @@ public function testUnsupportedRepo(): void
...
-     *   The expected exception.
-     * @param array $expectedUpdatedVersions
-     *   The expected repo file version after refresh attempt.
-     *
-     * @return void
-     *
-     * @dataProvider providerAttackRepoException
      */
-    public function testAttackRepoException(string $fixtureName, \Exception $expectedException, array $expectedUpdatedVersions): void
+    public function testRollbackAttackDetection(): void
```
phenaproxima, line 1244: This test case had only one scenario, and it was the same in both fixture variants. So I converted it to stop using a data provider, and renamed it for clarity in testdox.

tests/Metadata/FactoryTest.php
```diff
@@ -31,7 +31,7 @@ class FactoryTest extends TestCase
      */
     public function testLoad(string $class, string $role): void
     {
-        $clientStorage = static::loadFixtureIntoMemory('Delegated');
+        $clientStorage = static::loadFixtureIntoMemory('Delegated/consistent');
```
phenaproxima, line 34: The metadata tests are lower-level than the updater tests, so it made sense to have them continue to use the same exact fixtures (i.e., the consistent variants) they were using before.

tests/Metadata/MetaDataBaseTest.php
```diff
@@ -228,9 +228,27 @@ public function testMissingField(string $expectedField, string $exception = null
      */
     public function testOptionalFields(string $optionalField, $value): void
     {
+        $optionalField = explode(':', $optionalField);
+
         $metadata = json_decode($this->clientStorage[$this->validJson], true);
-        static::nestedChange(explode(':', $optionalField), $metadata, $value);
-        $json = json_encode($metadata);
+        static::nestedChange($optionalField, $metadata, $value);
+        $this->assertDataIsValid($metadata);
```
phenaproxima, line 235: Moved the assertion to its own method, since we need to use it more than once now.

tests/Metadata/RootMetadataTest.php
```diff
@@ -145,24 +145,13 @@ public function testInvalidRoleName(): void
...
+        foreach ([true, false] as $value) {
             $data['signed']['consistent_snapshot'] = $value;
             /** @var \Tuf\Metadata\RootMetadata $metadata */
             $metadata = static::callCreateFromJson(json_encode($data));
             $metadata->trust();
             $this->assertSame($value, $metadata->supportsConsistentSnapshots());
         }
-
-        // Ensure we get an error if consistent snapshots are not being used.
-        $data['signed']['consistent_snapshot'] = false;
```
phenaproxima, line 160: Removed this coverage because it is no longer accurate.

tests/TestHelpers/FixturesTrait.php
```diff
@@ -75,9 +75,9 @@ private static function loadFixtureIntoMemory(string $fixtureName, string $path
      * @return string
      *   The path.
      */
-    private static function getFixturePath(string $fixtureName, string $subPath = '', bool $isDir = true): string
+    protected static function getFixturePath(string $fixtureName, string $subPath = '', bool $isDir = true): string
```
phenaproxima, line 78: I think I needed to change this so it could be overridden in UpdaterTest.

phenaproxima (commented):
Just a self-review.

tedbow (approved):
Looks good!

PR #279, opened 2023-01-19
Remove TUFTestFixture prefix from all test fixtures

phenaproxima:
It's redundant, and reduces readability.


tedbow (approved):
looks good

PR #280, opened 2023-01-19
Encode fixtures' client start versions in dedicated files

phenaproxima:
Now that all fixtures have consistent and non-consistent versions, FixturesTrait::$initialMetadataVersions is too unwieldy. (Especially since the start versions will differ between the consistent and non-consistent versions.) We should move all that data to a specific file in each fixture.

For the moment, I'm using the `ini` format because, for our purposes:

* It's as readable as YAML.
* It supports comments.
* PHP has a built-in function to parse them (`parse_ini_file`), which means we don't need new dependencies. (However, I'm open to using YAML if we add a dev dependency on `ext-yaml`.)


tedbow (approved):
checked all the versions and they are the same.

looks good

PR #282, opened 2023-01-20
Encode both pre- and post-update versions in fixture-specific files

phenaproxima:
As I work to get #278 where it needs to be, I think it's becoming clear that we should double down on the concept introduced in #280, where the expected versions are encoded in a readable, comment-able file that's stored directly in the fixture.

The main benefit I think this will have is clarity. When you point a test case at a fixture, it will immediately know where it should start from, and where it should end up (assuming TUF doesn't throw an exception or abort the update).

@tedbow felt that the files should use YAML, and that's fine with me. It will necessitate adding a dev dependency on the Symfony YAML component (the YAML extension doesn't seem to install properly with PHP 8.2 on Windows).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -96,7 +96,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php }}
-          extensions: sodium, json, yaml
+          extensions: sodium, json
           tools: composer:v2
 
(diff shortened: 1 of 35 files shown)


phenaproxima:
On second thought, this won't work. Fixtures set initial conditions, but the final versions can vary depending on a number of factors. Those, then, should be encoded in the tests.

PR #283, opened 2023-01-23
Fix incorrect coverage annotations

phenaproxima:
Our coverage report (`composer coverage`) reports a lot of errors - that is, test methods which cover methods that don't exist. That's because our annotations aren't always correct.

This PR doesn't add or remove any coverage, but it does fix the wrong annotations so that PHPUnit can properly report what's covered and what's not.


TravisCarden (approved):
Everything looks good, @phenaproxima.

PR #284, opened 2023-01-25
Make metadata constraints easier to read by using default options for validation constraints

phenaproxima:
It'd be a little easier to read the validation constraints for metadata if we used the constraints' default options where possible. For example, things like `new Type(['type' => 'string'])` can always be converted to `new Type('string')`.

We already do this in several places, but I propose we make it consistent.

### DIFF
--- a/src/Metadata/ConstraintsTrait.php
+++ b/src/Metadata/ConstraintsTrait.php
@@ -135,7 +135,7 @@ trait ConstraintsTrait
             // is passed on from a setting in the Python `securesystemslib`
             // library.
             'keyid_hash_algorithms' => new Optional([
-                new EqualTo(["sha256", "sha512"]),
+                new EqualTo(['value' => ["sha256", "sha512"]]),
             ]),
             'keytype' => [
                 new Type('string'),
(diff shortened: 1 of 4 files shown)


src/Metadata/ConstraintsTrait.php
```diff
@@ -60,7 +60,7 @@ protected static function getHashesConstraints(): array
                 new Type('\ArrayObject'),
               // The keys for 'hashes is not know but they all must be strings.
                 new All([
-                    new Type(['type' => 'string']),
+                    new Type('string'),
```
tedbow, line 63:
```
public function __construct(string|array|null $type, string $message = null, array $groups = null, mixed $payload = 
    null, array $options = [])
    {
        if (\is_array($type) && \is_string(key($type))) {
            $options = array_merge($type, $options);
        } elseif (null !== $type) {
            $options['value'] = $type;
        }

        parent::__construct($options, $groups, $payload);

        $this->message = $message ?? $this->message;
    }
```
💩 wow, `string|array|null ` and docs 😜

but yes more readable on our side

tedbow (approved):
good, just 1 snarky comment

tedbow (changes requested):
I still see `new Type(['type' => 'string']),` that is not replaced in the codebase

PR #286, opened 2023-01-25
When downloading targets metadata, allow snapshot metadata to constrain the download size

phenaproxima:
According to the TUF spec, the snapshot can define _optional_ lengths for target metadata files: https://theupdateframework.github.io/specification/latest/#metapath-length.

> An integer length in bytes of the metadata file at [METAPATH](https://theupdateframework.github.io/specification/latest/#snapshot-metapath). It is OPTIONAL and can be omitted to reduce the snapshot metadata file size. In that case the client MUST use a custom download limit for the listed metadata.

Right now, we don't support this. We always (implicitly) constrain the download size of targets metadata, and any delegated roles, to Updater::MAXIMUM_DOWNLOAD_BYTES. Instead, we should default to that only if the snapshot doesn't define a length.

### DIFF
--- a/src/Client/Updater.php
+++ b/src/Client/Updater.php
@@ -469,16 +469,13 @@ class Updater
      */
     private function fetchAndVerifyTargetsMetadata(string $role): void
     {
-        /** @var RootMetadata $rootMetadata */
-        $rootMetadata = $this->storage->getRoot();
-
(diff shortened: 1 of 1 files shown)


phenaproxima:
This blocks #258.

phenaproxima:
Closing in favor of #258.

PR #287, opened 2023-01-27
Allow snapshot metadata to carry hashes for targets metadata

phenaproxima:
In the TUF spec, snapshot metadata can contain hashes for the targets metadata. Up to now, this hasn't been supported by us, but I think that for strong security it makes sense to do so.


tests/Client/UpdaterTest.php
```diff
@@ -998,15 +998,12 @@ public function providerExceptionForInvalidMetadata(): array
...
-            // For targets.json files, adding a new key or changing the existing version number
-            // will result in a SignatureThresholdException because currently the test
-            // fixtures do not contain hashes for targets.json files in snapshot.json.
             // § 5.6.3
             'add key to targets.json' => [
                 'targets.json',
                 ['signed', 'newvalue'],
                 'value',
-                new SignatureThresholdException("Signature threshold not met on targets"),
+                new MetadataException("The 'targets' contents does not match hash 'sha256' specified in the 'snapshot' metadata."),
```
tedbow, line 1006: Does mean we lose test coverage for `Signature threshold not met on targets`? couldn't still happen especially since `hashes` is optional. 

I search the code base for "Signature threshold not met on " and I only see 2 other roles that are hard-coded

tedbow (changes requested):
losing test coverage?

fixtures/Delegated/__init__.py
```diff
@@ -2,7 +2,7 @@
 
 
 def build():
-    fixture = ConsistencyVariantFixtureBuilder('Delegated')\
+    fixture = ConsistencyVariantFixtureBuilder('Delegated', { 'use_snapshot_hashes': True })\
```
tedbow, line 5: Do we need a fixture where this is false for testing?

phenaproxima: Not as far as I can tell. This fixture was meant for testing delegation of trust below the targets role, which isn't affected by snapshot and timestamp hasing.

tests/Client/UpdaterTest.php
phenaproxima: I reverted to the status of `main` and made better-scoped changes; this line is no longer affected by this PR.

tedbow (approved):
Looks good!

PR #289, opened 2023-02-01
Abstract server storage to increase mockability

phenaproxima:

PR #290, opened 2023-02-02
Improve separation of concerns

phenaproxima:
Supersedes #289.

I think we are sitting on a ticking time bomb of technical debt. There are a bunch of reasons for that, but one of them, in my opinion, is that we don't have the best separation of concerns in this library. It makes things somewhat inflexible and hard to test.

In this PR, I'd like to refactor the way PHP-TUF interacts with the server. There are really two aspects to that:

1. Loading data from the server.
2. Verify that it's safe enough to use.

These are true both when downloading TUF metadata, and actual targets. I think that we can greatly improve the situation by adopting the decorator pattern.

So I'm proposing we add `LoaderInterface`, which is a simple internal interface with one method, `load()`. This method returns a stream to an untrusted data source. To perform various safety checks on that data source, loaders (i.e., LoaderInterface implementations) should decorate other loaders. Example (which exists in this PR):

1. `GuzzleLoader` starts streaming data from some HTTP server.
2. `SizeCheckingLoader` wraps around that, and confirms that the stream never exceeds a given number of bytes.
3. `Updater` wraps around _that_, and confirms that the stream's contents match the hashes in the verified TUF metadata.

For tests, we can do something like this:

1. `FileLoader` loads files from the local file system (i.e., fixtures), as streams.
2. `SizeCheckingLoader` wraps around that, and checks the files aren't too big - same as it would do if loading data from a remote server.
3. `MaxBytesSpyLoader` wraps around _that_, recording arguments that its `load()` method was called with for later inspection.

To further improve the API, I also introduce `Repository`, which is a new class that accepts a loader, and exposes methods to fetch untrusted TUF metadata objects. It's just a cleaner way to get new metadata, and is an additional abstraction layer that we may be able to swap out later for testing purposes (although it's not swappable in this PR).

It's useful to isolate these concerns into specific classes, because it also makes testing them a lot easier. `SizeCheckingLoaderTest`, as an example, confirms that _no matter what_ a given stream is, or where it came from, it will always be length-checked in the way.

This stuff would replace `RepoFileFetcherInterface`, which is a bit more domain-specific and harder to work with. No other test coverage would be removed -- everything in UpdaterTest should still pass as normal.

I know this is a _major_ refactor, but I think it will help in the ultimate goal of making things cleaner and more testable.


composer.json
```diff
@@ -21,7 +21,6 @@
         "paragonie/sodium_compat": "^1.13",
         "symfony/validator": "^4.4 || ^5 || ^6",
         "myclabs/deep-copy": "^1.10.2",
-        "guzzlehttp/promises": "^1.5",
```
phenaproxima, line 24: In discussion with @tedbow, we agreed that we don't really need the potential asynchronicity offered by promises. So this PR removes promises support, and therefore we don't need this dependency anymore.

src/Client/Updater.php
```diff
@@ -24,17 +23,11 @@
  *
  * @package Tuf\Client
  */
-class Updater
+class Updater implements LoaderInterface
```
phenaproxima, line 26: Updater is a loader itself because it is, fundamentally, responsible for confirming that the contents of a stream conform to a known set of trusted hashes before passing that stream on. It was already doing this, but implementing LoaderInterface makes it official.

src/Client/Updater.php
```diff
@@ -24,17 +23,11 @@
...
 {
 
     const MAX_ROOT_DOWNLOADS = 1024;
 
-    /**
-     * The maximum number of bytes to download if the remote file size is not
-     * known.
-     */
-    const MAXIMUM_DOWNLOAD_BYTES = 100000;
```
phenaproxima, line 36: This constant makes more sense on Repository, which is responsible for loading metadata limited by size.

src/Client/Updater.php
```diff
@@ -375,8 +318,6 @@ protected function verify(string $target, StreamInterface $data): void
         if ($targetsMetadata === null) {
             throw new NotFoundException($target, 'Target');
         }
-        $maxBytes = $targetsMetadata->getLength($target) ?? static::MAXIMUM_DOWNLOAD_BYTES;
```
phenaproxima, line 378: There's no further need for this, because SizeCheckingLoader takes care of it before the stream even reaches this method.

src/Loader/LoaderInterface.php
```diff
@@ -0,0 +1,37 @@
...
+
+use Psr\Http\Message\StreamInterface;
+
+/**
+ * Defines an interface to load data as a stream.
+ *
+ * The data source can be anything, from anywhere, but it must be returned as an
+ * implementation of \Psr\Http\Message\StreamInterface.
+ *
+ * This is an internal interface used to load untrusted data! External code
```
phenaproxima, line 13: Right now, there's no real reason for anything external to TUF to implement this interface. There may be in the future, but for now it makes sense for it to be internal. It needs to exist as an interface in order for the decorator pattern to work properly.

src/Client/Updater.php
```diff
@@ -90,18 +69,28 @@ class Updater
...
+     *
+     *@todo What is this for?
+     *       https://github.com/php-tuf/php-tuf/issues/161
      */
-    public function __construct(RepoFileFetcherInterface $repoFileFetcher, StorageInterface $storage)
+    public function __construct(private SizeCheckingLoader $loader, protected StorageInterface $storage)
     {
-        $this->repoFileFetcher = $repoFileFetcher;
+        $this->server = new Repository($this->loader);
         $this->storage = $storage;
```
tedbow, line 94: We no longer have this property

src/Loader/LoaderInterface.php
tedbow: I am not sure that Updater should implement this interface. Does `\Tuf\Client\Updater::load` return untrusted data?

src/Client/Repository.php
```diff
@@ -0,0 +1,108 @@
...
+     *   The version of the root metadata to load.
+     *
+     * @return \Tuf\Metadata\RootMetadata|null
+     *   An instance of \Tuf\Metadata\RootMetadata, or null if the requested
+     *   version of the metadata doesn't exist.
+     */
+    public function getRoot(int $version): ?RootMetadata
+    {
+        try {
+            $data = $this->loader->load("$version.root.json", self::MAX_BYTES);
```
tedbow, line 42: It seems like if the class property wasn't named `loader` generically it would be clear what is going on here. could

src/Loader/SizeCheckingLoader.php
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function load(string $uri, int $maxBytes = null): StreamInterface
+    {
+        $data = $this->decorated->load($uri, $maxBytes);
+
+        if (isset($maxBytes)) {
```
tedbow, line 24: It seems like that fact that this could be NULL defeats the whole purpose. 

Could we enforce it being not NULL?

src/Client/Repository.php
phenaproxima: How 'bout `dataLoader` as a name for the class property?

src/Client/Updater.php
phenaproxima: Well-spotted. Fixed.

src/Client/Repository.php
phenaproxima: Renamed to sizeCheckingLoader per discussion on Zoom.

src/Loader/SizeCheckingLoader.php
phenaproxima: After discussion on Zoom, maxBytes is now required by the interface.

src/Loader/LoaderInterface.php
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Loads data as a stream.
+     *
+     * @param string $uri
+     *   The URI of the data to load. The meaning of this depends on the
+     *   implementing class; it could be a URL, a relative or absolute file
+     *   path, or something else.
+     * @param int $maxBytes
+     *   (optional) The maximum number of bytes that should be read from the
+     *   data source, or null for no limit.
```
tedbow, line 28: not optional now

src/Loader/LoaderInterface.php
```diff
@@ -0,0 +1,37 @@
...
+ */
+interface LoaderInterface
+{
+    /**
+     * Loads data as a stream.
+     *
+     * @param string $uri
+     *   The URI of the data to load. The meaning of this depends on the
+     *   implementing class; it could be a URL, a relative or absolute file
+     *   path, or something else.
```
tedbow, line 25: I don't think a `URI` is right name for this.  URI is does a definition and I don't think a "relative file path" fall into this this. could we change this to "locator" especially since we include "something else" as a possible value

https://stackoverflow.com/questions/44692872/is-a-local-file-path-a-uri

tests/Client/TestLoader.php
```diff
@@ -33,7 +28,7 @@ class TestRepo implements RepoFileFetcherInterface
      *
      * @var array[]
      */
-    public array $fetchMetadataArguments = [];
+    public array $maxBytes = [];
```
tedbow, line 31: The doc for this property needs to be updated

src/Client/Updater.php
```diff
@@ -391,34 +316,27 @@ protected function verify(string $target, StreamInterface $data): void
...
+     *   The maximum number of bytes to download, or null to have no limit.
      *
-     * @return \GuzzleHttp\Promise\PromiseInterface
-     *   A promise representing the eventual verified result of the download
-     *   operation.
+     * @return \Psr\Http\Message\StreamInterface
+     *   A stream of the trusted, downloaded data.
      */
-    public function download(string $target, ...$extra): PromiseInterface
+    public function download(string $target, int $maxBytes = null): StreamInterface
```
tedbow, line 325: `$maxBytes` is unused

src/Loader/LoaderInterface.php
phenaproxima: Very true. Fixed!

tests/Client/TestLoader.php
phenaproxima: Fixed.

src/Client/Updater.php
phenaproxima: Indeed it is. Fixed.

src/Loader/LoaderInterface.php
phenaproxima: That's a good point. Removed the interface from Updater.

tedbow (approved):
Looks good

PR #296, opened 2023-02-07
Support non-seekable streams in SizeCheckingLoader

### LINKED ISSUE #169: Handle non-seekable streams in Updater:: checkLength()
Coming at you from #153!

There's an outstanding @todo at https://github.com/php-tuf/php-tuf/blob/main/src/Client/Updater.php#L535 which states:

```
        if (isset($size)) {
            if ($size > $maxBytes) {
                throw $error;
            }
        } else {
            // @todo Handle non-seekable streams.
            $data->rewind();
            $data->read($maxBytes);
```

Seems like a good thing to do. :)

phenaproxima:
Fixes #169.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,6 @@
         "slevomat/coding-standard": "^8.2",
         "phpspec/prophecy-phpunit": "^2",
         "guzzlehttp/guzzle": "^6.5 || ^7.2",
-        "guzzlehttp/psr7": "^1.7",
         "phpspec/prophecy": "^1.16"
     },
     "license": "MIT",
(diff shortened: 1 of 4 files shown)


phenaproxima:
Due to the fact that this intentionally introduces side effects, I think I'm going to want sign-off from both @tedbow and @effulgentsia before this is merged.

tedbow (approved):
Looks good!

phenaproxima:
I confirmed that the changes here un-break the tests in php-tuf/composer-integration.

phenaproxima:
Welp, since this blocks the plugin, I'm mergin' this.

PR #297, opened 2023-02-07
Remove GuzzleLoader

phenaproxima:
Now that https://github.com/php-tuf/composer-integration/pull/77 is in, nothing is using GuzzleLoader. So let's remove it!


tedbow (approved):
👍

PR #299, opened 2023-02-08
Add explicit test coverage for unsupported key types

### LINKED ISSUE #168: Validate key type of $signatureMeta in Updater::verifySingleSignature()
Coming at you from #153!

There's an outstanding @todo at https://github.com/php-tuf/php-tuf/blob/main/src/Client/Updater.php#L389 which states:

```
        // Encode the pubkey and signature, and check that the signature is
        // valid for the given data and pubkey.
        $pubkeyBytes = hex2bin($pubkey);
        $sigBytes = hex2bin($signatureMeta['sig']);
        // @todo Check that the key type in $signatureMeta is ed25519; return
        //     false if not.
        return \sodium_crypto_sign_verify_detached($sigBytes, $bytes, $pubkeyBytes);
```

Let's do it!

phenaproxima:
Fixes #168 by adding test coverage.

### DIFF
--- a/src/KeyDB.php
+++ b/src/KeyDB.php
@@ -88,10 +88,11 @@ class KeyDB
      */
     public function addKey(string $keyId, Key $key): void
     {
-        if (! in_array($key->getType(), self::getSupportedKeyTypes(), true)) {
+        $keyType = $key->getType();
+        if (! in_array($keyType, self::getSupportedKeyTypes(), true)) {
             // @todo Convert this to a log line as per Python.
             // https://github.com/php-tuf/php-tuf/issues/160
-            throw new InvalidKeyException("Root metadata file contains an unsupported key type: \"{$keyMeta['keytype']}\"");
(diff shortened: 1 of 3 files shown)


src/KeyDB.php
```diff
@@ -88,10 +88,11 @@ public function __construct()
      */
     public function addKey(string $keyId, Key $key): void
     {
-        if (! in_array($key->getType(), self::getSupportedKeyTypes(), true)) {
+        $keyType = $key->getType();
+        if (! in_array($keyType, self::getSupportedKeyTypes(), true)) {
             // @todo Convert this to a log line as per Python.
             // https://github.com/php-tuf/php-tuf/issues/160
-            throw new InvalidKeyException("Root metadata file contains an unsupported key type: \"{$keyMeta['keytype']}\"");
```
phenaproxima, line 94: $keyMeta is an undefined variable. The fact that this never caused a test failure is proof positive that this code has never been exercised in our test cases until now.

tedbow (approved):

PR #300, opened 2023-02-09
Require guzzlehttp/psr7 to 2.4

phenaproxima:
We currently use guzzlehttp/psr7 v1, but it's not compatible with Drupal 10, which requires 2.4 or later. I don't see any real reason not to bump our constraint.


tedbow (approved):
looks fine if the tests pass

PR #304, opened 2023-02-14
Require only supported versions of symfony/validator

phenaproxima:
According to https://endoflife.date/symfony, all Symfony 5 versions before 5.4 and Symfony 6 versions before 6.2 are EOL. We should bump our constraints to prevent older versions from being accidentally installed.


phenaproxima:
As of #362, we only support Symfony 6, which is supported. This is obsolete, so closing it out.

PR #306, opened 2023-02-14
Remove direct dependency on psr/http-message

phenaproxima:
I'm not convinced we need this as a direct dependency.

We already have a hard dependency on guzzlehttp/psr7, which is by definition an implementation of the interfaces in psr/http-message. So it seems redundant to also depend on psr/http-message. I propose we remove it until -- and if -- we remove guzzlehttp/psr7.


tedbow (approved):
👍

PR #307, opened 2023-02-14
Remove guzzlehttp/guzzle from composer.json suggestions

phenaproxima:
So...since #297, we no longer have any use of the Guzzle HTTP client. So, the suggestion in composer.json to install it is not accurate anymore and should be removed.


tedbow (approved):
👍

PR #308, opened 2023-02-14
Remove ResponseStream

phenaproxima:
This class is not used by PHP-TUF or the Composer plugin. It's also untested.

I think that this is a relic from a time when we needed access to an untrusted response for some reason, but if we're not using it anywhere, then I guess that time has passed. We should remove the class.


tedbow (approved):
good find!

PR #309, opened 2023-03-21
Adopt typed properties and constructor property promotion

phenaproxima:
Because it's 2023, and our minimum supported PHP is 8.0. This removes a pile of annoying boilerplate, and anything that increases type safety is good.

### DIFF
--- a/src/KeyDB.php
+++ b/src/KeyDB.php
@@ -22,6 +22,13 @@ use Tuf\Metadata\RootMetadata;
 class KeyDB
 {
     /**
+     * A set of keys, indexed by key ID.
+     *
+     * @var array[]
+     */
+    protected array $keys = [];
+
+    /**
      * Creates a key database with the unique keys found in root metadata.
      *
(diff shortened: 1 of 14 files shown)


src/KeyDB.php
```diff
@@ -68,10 +61,12 @@ public static function getSupportedKeyTypes(): array
 
     /**
      * Constructs a new KeyDB.
+     *
+     * @param array $keys
+     *   Keys indexed by key ID.
      */
-    public function __construct()
+    public function __construct(protected array $keys = [])
```
tedbow, line 68: Why do we want this to be setable via constructor now?

Please on comment on either other changes link this and then re-assign to me.

tedbow (changes requested):
Need comments on un-related / unneeded changes

src/KeyDB.php
phenaproxima: Good point. Must have missed that. Reverted (and removed the constructor entirely).

tedbow (approved):
Looks good!

PR #311, opened 2023-03-22
Update TUF spec to 1.0.19

phenaproxima:
See https://github.com/theupdateframework/specification/compare/v1.0.18...v1.0.19.

### DIFF
--- a/src/Client/Updater.php
+++ b/src/Client/Updater.php
@@ -434,7 +434,7 @@ class Updater
                     return $delegatedTargetsMetadata;
                 }
                 $searchedRoles[] = $delegatedRoleName;
-                // § 5.6.7.2.1
+                // § 5.6.7.2
                 // Recursively search the list of delegations in order of appearance.
                 $delegatedRolesMetadataSearchResult = $this->searchDelegatedRolesForTarget($delegatedTargetsMetadata, $target, $searchedRoles, $terminated);
                 if ($terminated || $delegatedRolesMetadataSearchResult) {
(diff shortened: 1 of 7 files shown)


src/Client/Updater.php
```diff
@@ -445,7 +445,7 @@ private function searchDelegatedRolesForTarget(TargetsMetadata $targetsMetadata,
                 // in the delegations from $targetsMetadata.
                 if ($delegatedRole->isTerminating()) {
                     $terminated = true;
-                    // § 5.6.7.2.2
+                    // § 5.6.7.2.1
```
tedbow, line 448: There still some lines in this branch that refer to `5.6.7.2.2` that have not been changed

tedbow (changes requested):
Just some things to replace

src/Client/Updater.php
phenaproxima: Thanks for pointing that out.

It looks like most of the references are actually still valid. But I did remove a reference to 5.6.7.2.3, which no longer exists.

tedbow: I see lines like "`// See § 5.6.7.2.1 and 5.6.7.2.2.`"

since 2.1 and 2.2 are not different. Then this now refers to different points in .19. Were they wrong before?

phenaproxima: My interpretation is that they weren't wrong; they were merely very general. I think it was our way of saying "we are dealing with delegation stuff".

I suspect they were wrong before, in a subtle way. The only point in Updater that mentioned 5.6.7.2.1 had this comment below it: "Recursively search the list of delegations in order of appearance." There was, and is, no apparent support for multi-role delegations (whatever those are).

tedbow (approved):
Looks good

PR #312, opened 2023-03-22
Update TUF spec to v1.0.20

phenaproxima:
See https://github.com/theupdateframework/specification/compare/v1.0.19...v1.0.20.

### DIFF
--- a/.github/workflows/specification-version-check.yml
+++ b/.github/workflows/specification-version-check.yml
@@ -11,4 +11,4 @@ jobs:
       issues: write
     uses: theupdateframework/specification/.github/workflows/check-latest-spec-version.yml@master
     with:
-      tuf-version: "v1.0.19" # Should be updated to the version the project supports either manually or extracted automatically. You can see how python-tuf did that as an example.
+      tuf-version: "v1.0.20" # Should be updated to the version the project supports either manually or extracted automatically. You can see how python-tuf did that as an example.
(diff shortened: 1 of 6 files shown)


phenaproxima:
The main difference in this version seems to be a clarification of path patterns and what they match. As far as I know, `fnmatch()` (which we use to do this comparison) is compliant, since [its documentation says it matches shell-style wildcards](https://www.php.net/fnmatch), which [matches the spec](https://github.com/theupdateframework/specification/compare/v1.0.19...v1.0.20#diff-b0020aab85257910a5a31f248c201a81cb35e1939545c4b45515e9049e78cdffR1044).

This spec version also clarifies that delegated targets must match the path patterns of all roles in the delegation chain, but that is something we already support and have test coverage for (see `\Tuf\Tests\Client\UpdaterTest::providerDelegationErrors()`).

tedbow:
I see in https://github.com/theupdateframework/specification/compare/v1.0.19...v1.0.20#diff-b0020aab85257910a5a31f248c201a81cb35e1939545c4b45515e9049e78cdffR1050

`it is RECOMMENDED that <a>PATHPATTERN</a> uses the forward slash (`/`) as
    directory separator and does not start with a directory separator, as is
    also recommended for <a>TARGETPATH</a>. A path separator in a path SHOULD
    NOT be matched by a wildcard in the <a>PATHPATTERN</a>.`

but if I look at `\Tuf\Tests\Unit\DelegatedRoleTest::providerMatchesRolePath` we are having our test paths "start with a directory separator. Should we change this?

PR #314, opened 2023-03-22
Update TUF spec version to 1.0.21

phenaproxima:
See https://github.com/theupdateframework/specification/compare/v1.0.20...v1.0.21.

The differences are that key IDs and role names are now required to be unique, which means we should do some enforcement.

We can't really use Symfony's `Unique` constraint for this, because we'd need to use the `fields` option, which was only added in Symfony 6.1, which requires PHP 8.1 at a minimum. So we'll just do custom validation instead, and leave the discussion of dropping support for Symfony 4 and 5, as well as PHP 8.0, for another time.


src/Metadata/MetadataBase.php
```diff
@@ -133,6 +136,22 @@ protected static function getConstraints(): array
...
+    /**
+     * {@inheritdoc}
+     */
+    protected static function validate(array $data, Collection $constraints): void
+    {
+        static::traitValidate($data, $constraints);
+
+        // The TUF spec requires that all key IDs be unique.
+        // @todo Use Symfony's Unique constraint for this when at least Symfony
+        //   6.1 is required.
```
tedbow, line 148: Can we make an issue for this to avoid adding another `todo` without a follow-up issue

phenaproxima: Opened #317 for that.

src/Metadata/MetadataBase.php
```diff
@@ -133,6 +136,22 @@ protected static function getConstraints(): array
...
+        $keyIds = array_column($data['signatures'], 'keyid');
```
tedbow, line 149: @phenaproxima did you consider using `\Symfony\Component\Validator\Constraints\Callback` to a static method so to the we could keep all the config for the metadata validation  in one place. Depending on if we get any usage from other php projects it could be awhile before we can drop support for symfony 5 as it security support till 2025 if this is correct https://endoflife.date/symfony

phenaproxima: 🤔 I didn't even know there was such a constraint.

tedbow (changes requested):
Just the 1 question about the callback constraint

and we need a link to the new isseu https://github.com/php-tuf/php-tuf/issues/317 in the current todo

src/Metadata/MetadataBase.php
phenaproxima: I changed it to use the Callback constraint. Honestly it's not a huge difference because the callback has to be static, but it's a nice clean-up. I could make it a static closure if you'd prefer.

src/Metadata/TargetsMetadata.php
```diff
@@ -42,6 +44,28 @@ public static function createFromJson(string $json, string $roleName = null): st
...
+     *   6.1 is required in https://github.com/php-tuf/php-tuf/issues/317.
+     *
+     * @param mixed $delegations
+     *   The value to be validated.
+     * @param \Symfony\Component\Validator\Context\ExecutionContextInterface $context
+     *   The validation context.
+     */
+    public static function validateDelegatedRoles(mixed $delegations, ExecutionContextInterface $context): void
+    {
+        if (!is_array($delegations)) {
```
tedbow, line 60: Would it possible to add the validation on the `'roles' => new All([` level so you would not have to check if this was set?

tedbow (approved):
Looks good!

src/Metadata/TargetsMetadata.php
phenaproxima: According to https://symfony.com/doc/current/reference/constraints/All.html, that's not what `All` does.

The reason I had to add this check, for posterity, is because type hinting `$delegations` as an array causes some tests to fail -- specifically, `MetadataBaseTest::testInvalidFields()`, which specifically passes the wrong data type in, so as to confirm we're testing data types. That test coverage is valuable, but fundamentally incompatible with adding a type hint here, because the test will start failing with `TypeError`.

PR #315, opened 2023-03-23
Remove the need for DeepCopy and ArrayObject, and allow metadata objects to do custom normalization

phenaproxima:
We have a ton of complexity surrounding our normalization stuff, introduced in #88. And we can't really properly update to TUF spec v1.0.21 unless we untangle it.

After researching the alternatives, I think the clearest and easiest way to do this is...to stop using ArrayObject as a workaround for normalization issues. We can instead have `MetadataBase` implement PHP's `JsonSerializable` interface, and then cast empty arrays to objects as needed. (As it turns out, the only one that really needs to do this is `TargetsMetadata`.) Any normalization we need to do takes place in `jsonSerialize()`, and we can explicitly document those decisions in comments, rather than having to deal with ArrayObjects for unclear reasons.

This _completely_ removes a lot of complexity, as well as the need for the DeepCopy dependency. Arrays automatically copy by value, so there's no chance of memory shenanigans or security/memory problems coming from upstream. The fact that we don't need to deep copy anymore is probably beneficial for performance as well. Arrays are a little easier to debug too.


phenaproxima:
This blocks #314.

src/Client/SignatureVerifier.php
```diff
@@ -58,7 +57,7 @@ public function checkSignatures(MetadataBase $metadata): void
         $needVerified = $role->getThreshold();
         $verifiedKeySignatures = [];
 
-        $canonicalBytes = JsonNormalizer::asNormalizedJson($metadata->getSigned());
```
tedbow, line 61: So the only non-test use I see of `\Tuf\JsonNormalizer::asNormalizedJson. Could we just do:

`$keyCanonicalForm = json_encode(JsonNormalizer::rKeySort($keyCanonicalStruct), JSON_UNESCAPED_SLASHES);`
or something and just remove asNormalizedJson

src/Metadata/TargetsMetadata.php
```diff
@@ -42,6 +42,29 @@ public static function createFromJson(string $json, string $roleName = null): st
...
+    {
+        $signedData = parent::jsonSerialize();
+
+        foreach ($signedData['targets'] as $path => $target) {
+            // Custom target info should always encode to an object, even if
+            // it's empty.
+            if (array_key_exists('custom', $target)) {
+                $target['custom'] = (object) $target['custom'];
+            }
+            $signedData['targets'][$path] = $target;
```
tedbow, line 58: couldn't this be inside the if?

tedbow (changes requested):

src/Client/SignatureVerifier.php
phenaproxima: Yup, I was considering something like this. Thought it might be out of scope, but if you're +1, then I can do that.

tedbow: sure

phenaproxima: OK, I've created a new `CanonicalJsonTrait` which is the successor of `JsonNormalizer`. Metadata objects now have a `toCanonicalJson()` method.

src/Metadata/TargetsMetadata.php
phenaproxima: Sure. Fixed.

src/CanonicalJsonTrait.php
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * Sorts an associative array into a canonical order.
+     *
+     * @param array $data
+     *     The data to sort, passed by reference.
+     *
+     * @throws \RuntimeException
+     *     Thrown if sorting the array fails.
+     */
+    protected static function sortKeys(array &$data): void
```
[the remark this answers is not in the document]
phenaproxima: Sure, although it won't matter in a trait. Fixed.

src/Role.php
```diff
@@ -37,7 +37,7 @@ protected function __construct(protected string $name, protected int $threshold,
      *
      * @see https://theupdateframework.github.io/specification/v1.0.20#document-formats
      */
-    public static function createFromMetadata(\ArrayObject $roleInfo, string $name): Role
+    public static function createFromMetadata(iterable $roleInfo, string $name): Role
```
tedbow, line 40: `@param \ArrayObject $roleInfo` still needs to be updated. PHPCS doesn't catch this?

phenaproxima: Fixed. I guess PHPStorm didn't catch it because the actual type hint is `iterable`, which `ArrayObject` falls into. I've changed them both to `array`.

src/Client/SignatureVerifier.php
```diff
@@ -93,7 +91,7 @@ public function checkSignatures(MetadataBase $metadata): void
      * @return boolean
      *     TRUE if the signature is valid for $bytes.
      */
-    private function verifySingleSignature(string $bytes, \ArrayAccess $signatureMeta): bool
+    private function verifySingleSignature(string $bytes, array $signatureMeta): bool
```
tedbow, line 94: `@param` needs to be updated

tedbow (approved):
Looks good!

@phenaproxima should phpcs have caught the `@param`'s that were not updated or is that phpstan. In any case should we make follow-up to implement that? Would make reviews and maintainability easier

PR #319, opened 2023-03-24
Update TUF spec to v1.0.23

phenaproxima:
Behold: https://github.com/theupdateframework/specification/compare/v1.0.22...v1.0.23

As far as I can tell, we are already compliant with this. When computing hashes and whatnot, we simply do it by canonicalizing everything in `signed`. We don't care what's in there.


tedbow:
@phenaproxima I feel like we had test coverage extra fields being ignore. Do you know if that is true?

phenaproxima:
Looks like we do. See `\Tuf\Tests\Metadata\TargetsMetadataTest::providerOptionalFields()`.

tedbow (approved):
Looks good

PR #320, opened 2023-03-24
Update TUF spec to v1.0.24

phenaproxima:
See https://github.com/theupdateframework/specification/compare/v1.0.23...v1.0.24

Three relevant bits here:

> An OPTIONAL boolean indicating whether the repository supports consistent snapshots.

This is already optional, and tested in `\Tuf\Tests\Metadata\RootMetadataTest::providerOptionalFields()`.

> DELEGATIONS is an OPTIONAL object

This is also already the case, and tested by `\Tuf\Tests\Metadata\TargetsMetadataTest::providerOptionalFields()`.

> The "path_hash_prefixes" and "paths" attributes are OPTIONAL

Ah, _this_ we needed to change. I did that, and added test coverage in TargetsMetadataTest. I confirmed that removing the `Optional` constraint from `paths` causes the test to fail.

### DIFF
--- a/.github/workflows/specification-version-check.yml
+++ b/.github/workflows/specification-version-check.yml
@@ -11,4 +11,4 @@ jobs:
       issues: write
     uses: theupdateframework/specification/.github/workflows/check-latest-spec-version.yml@master
     with:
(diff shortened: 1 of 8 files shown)


tedbow (approved):
Great!

PR #321, opened 2023-03-24
Update TUF spec to v1.0.25

phenaproxima:
See https://github.com/theupdateframework/specification/compare/v1.0.24...v1.0.25

We already already compliant with this (see, for example, `\Tuf\Metadata\Verifier\SnapshotVerifier::verify()`). We check hashes first, and that seems to be what the correct behavior is, but I think they're just clarifying _why_ it's safe to check the hashes before the signatures.


tedbow (approved):
yep, looks good

PR #322, opened 2023-03-24
Update TUF spec to v1.0.26

phenaproxima:
The diff: https://github.com/theupdateframework/specification/compare/v1.0.25...v1.0.26

Although this looks intimidating, this update is all about the _server_ side and how it writes metadata. I don't think there are any changes, per se; it's just that the language is stricter, clearer, and more formal. There are some clarifications. We're already consistent with the described behavior.


tedbow (approved):
Yep looks like this not for client code

PR #326, opened 2023-03-27
Update TUF spec to v1.0.30

phenaproxima:
See https://github.com/theupdateframework/specification/compare/v1.0.29...v1.0.30

This bump actually has some behavior changes that need a little refactoring on our part. In particular, the behavior of root metadata verification has changed; we need a way to continue the update process using the current trusted metadata. I also think we need to distinguish between rollback attacks ("this new version is older than what we expected") versus mix-and-match attacks ("this version is not what we expected"); the two are conflated right now.


src/Client/Updater.php
```diff
@@ -167,12 +171,14 @@ public function refresh(bool $force = false): bool
...
-    private function updateTimestamp(): TimestampMetadata
+    private function updateTimestamp(): ?TimestampMetadata
     {
         // § 5.4.1
-        $newTimestampData = $this->server->getTimestamp();
-
-        $this->universalVerifier->verify(TimestampMetadata::TYPE, $newTimestampData);
+        /** @var \Tuf\Metadata\TimestampMetadata $newTimestampData */
+        $newTimestampData = $this->universalVerifier->verify(TimestampMetadata::TYPE, $this->server->getTimestamp());
+        if ($newTimestampData === $this->storage->getTimestamp()) {
```
tedbow, line 179: Looking at https://github.com/theupdateframework/specification/pull/209 which is one of the commits for the updated tag

> timestamp should be updated only when there's a newer version AND the referenced snapshot version is not less than the one in the already trusted timestamp

Are we comparing the versions of the snapshot referenced in the new and existing timestamps?

tedbow (approved):
Looks good!

PR #327, opened 2023-03-27
Use validation constraints to check key type

phenaproxima:
This is some clean-up in the name of brevity. Right now, we do a specific check in KeyDB::addKey() that the incoming key is of a supported type. Given how KeyDB is created (basically it's just pulling directly from some root metadata), there is no reason why we shouldn't just use a validation constraint for this. It's simpler and lets us keep related validation logic in one place.

### DIFF
--- a/src/KeyDB.php
+++ b/src/KeyDB.php
@@ -56,18 +56,6 @@ class KeyDB
     }
 
     /**
-     * Gets the supported encryption key types.
-     *
-     * @return string[]
-     *     An array of supported encryption key type names (e.g. 'ed25519').
-     */
-    public static function getSupportedKeyTypes(): array
-    {
-        return ['ed25519'];
-    }
-
-
-    /**
      * Adds key metadata to the key database while avoiding duplicates.
      *
      * @param string $keyId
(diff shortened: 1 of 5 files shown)


src/KeyDB.php
```diff
@@ -55,18 +55,6 @@ public static function createFromRootMetadata(RootMetadata $rootMetadata, bool $
         return $db;
     }
 
-    /**
-     * Gets the supported encryption key types.
-     *
-     * @return string[]
-     *     An array of supported encryption key type names (e.g. 'ed25519').
-     */
-    public static function getSupportedKeyTypes(): array
```
tedbow, line 64: Was there any reason this was public before? If are going to put into validation should we just add note to the readme saying we just support this 1 type?

tedbow (changes requested):
generally looks good. 1 suggestion

src/KeyDB.php
phenaproxima: After discussion, I agreed to mention this in in the wiki, since it's not very discoverable otherwise. https://github.com/php-tuf/php-tuf/wiki#notes

tedbow (approved):
👍

PR #328, opened 2023-03-27
Use sodium_hex2bin

phenaproxima:
In SignatureVerifier, we currently call PHP's `hex2bin()`. According to [PHP's documentation](https://www.php.net/sodium_hex2bin), this function is susceptible to side-channel attacks, whereas `sodium_hex2bin` isn't. Since we already use sodium functions, we should use `sodium_hex2bin` here.


tedbow (approved):
👍

PR #329, opened 2023-03-27
Remove KeyDB and merge its functionality into SignatureVerifier

phenaproxima:
I'm not convinced we need KeyDB.

The only thing that really needs to collect keys is the signature verifier; it has redundant code that wraps pointlessly around KeyDB. I think we can simplify this by just having SignatureVerifier implement KeyDB's functionality.


tests/SignatureVerifierTest.php
```diff
@@ -31,12 +30,17 @@ public function testCreateFromRootMetadata(): void
...
-        self::assertInstanceOf(KeyDB::class, $keyDb);
+
+        $verifier = SignatureVerifier::createFromRootMetadata($rootMetadata);
+        $verifierKeys = new \ReflectionProperty($verifier, 'keys');
+        $verifierKeys->setAccessible(true);
+        $verifierKeys = $verifierKeys->getValue($verifier);
+
         // Get the first key for comparison.
         $keys = $rootMetadata->getKeys();
         $key = reset($keys);
```
tedbow, line 41: Why aren't we just looping through all the keys?

phenaproxima: We could absolutely do that. But this was the existing test coverage from KeyDBTest, so I decided not to alter it (so that you would approve it without asking questions :))

tedbow (approved):
👍

PR #333, opened 2023-03-29
In preparation for breaking up UpdaterTest, create a base class for functionally testing the updater

phenaproxima:
UpdaterTest is, to be frank, an incomprehensible mess. It's tightly coupled to the fixtures in ways that make it very hard to modify. It needs to be heavily refactored, and that's gonna be a bit of a project. The first step is to prepare to break it up into smaller tests.

To facilitate that, I think we should create a ClientTestBase which implements the stuff needed to functionally test the updater: namely, the ability to load fixture files into client and server storage, and modify those files, separately from setting up the test updater itself. To prove the concept and increased readability, this PR also splits some of the simpler tests into their own classes.


tests/TestHelpers/FixturesTrait.php
```diff
@@ -78,29 +78,18 @@ protected static function getFixturePath(string $fixtureName, string $subPath =
...
-                    break;
-                case 'snapshot':
-                    $metadata = $storage->getSnapshot();
-                    break;
-                default:
-                    // Any other roles will be 'targets' or delegated targets roles.
-                    $metadata = $storage->getTargets($role);
-                    break;
-            }
+            $metadata = match ($role) {
```
phenaproxima, line 87: This change is to take advantage of the `match` expression's exhaustiveness and strict type checking.

tests/TestHelpers/FixturesTrait.php
```diff
@@ -110,7 +99,7 @@ private static function assertMetadataVersions(array $expectedVersions, StorageI
 
             $actualVersion = $metadata->getVersion();
             Assert::assertSame(
-                $expectedVersions[$role],
+                $version,
```
phenaproxima, line 102: Just a redundant variable.

tests/Client/RoleDownloadLimitTest.php
```diff
@@ -0,0 +1,51 @@
...
+        self::assertNotEmpty($testFileContents);
+        self::assertSame($testFileContents, $this->getUpdater()->download($fileName)->getContents());
+
+        // Ensure the file can not found if the maximum role limit is 3.
+        self::expectException(NotFoundException::class);
+        self::expectExceptionMessage("Target not found: $fileName");
+        $this->getUpdater(LimitRolesTestUpdater::class)->download($fileName);
+    }
+}
+// @codingStandardsIgnoreStart
```
phenaproxima, line 39: Coding standards are disabled in this section because they want every class to be in its own file. But IMHO it is clearer to have this one class in this file, since this is the only place it's ever used.

tests/Client/InconsistentFixturesUpdaterTest.php
```diff
@@ -113,14 +115,4 @@ public function providerFileNotFoundExceptions(): array
         $data['targets.json in Delegated'][2]['root'] = 3;
         return $data;
     }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function providerUnsupportedRepo(): array
-    {
```
phenaproxima, line 121: This was moved into the data provider for `PathHashPrefixesTest`.

tests/Client/UpdaterTest.php
```diff
@@ -2,148 +2,25 @@
...
-    }
-
-    /**
-     * Tests that TUF will transparently verify downloaded target hashes.
-     *
-     * @covers ::download
-     *
-     * @return void
-     */
-    public function testVerifiedDownload(): void
```
phenaproxima, line 110: This was moved into `VerifiedDownloadTest`.

tests/Client/UpdaterTest.php
```diff
@@ -1109,146 +949,34 @@ public function providerFileNotFoundExceptions(): array
...
-
-    /**
-     * Data provider for testSignatureThresholds().
-     *
-     * @return mixed[]
-     *   The test cases for testSignatureThresholds().
-     */
-    public function providerTestSignatureThresholds():array
-    {
-        return [
```
phenaproxima, line 1121: This was converted to a `@testWith` annotation.

tests/Client/UpdaterTest.php
```diff
@@ -1109,146 +949,34 @@ public function providerFileNotFoundExceptions(): array
...
 
             $this->expectException(SignatureThresholdException::class);
         }
-        $updater->refresh();
-    }
-
-    /**
-     * Tests forcing a refresh when the server is in an invalid state.
-     */
-    public function testRefreshFromServerInInvalidState(): void
```
phenaproxima, line 1166: This is now `InvalidRefreshTest`.

tests/Client/UpdaterTest.php
```diff
@@ -1109,146 +949,34 @@ public function providerFileNotFoundExceptions(): array
...
-    }
-
-    /**
-     * Tests that an exceptions for an repo with an unsupported field.
-     *
-     * @dataProvider providerUnsupportedRepo
-     */
-    public function testUnsupportedRepo(array $expectedUpdatedVersion): void
-    {
-        $fixtureSet = 'UnsupportedDelegation';
```
phenaproxima, line 1208: This method, along with its data provider, is now in `PathHashPrefixesTest`.

tests/Client/UpdaterTest.php
```diff
@@ -1109,146 +949,34 @@ public function providerFileNotFoundExceptions(): array
...
-            self::assertSame($this->clientStorage->read('targets'), $startingTargets);
-            return;
-        }
-        $this->fail('No exception thrown.');
-    }
-
-    /**
-     * Tests that exceptions are thrown when a repo is in a rollback attack state.
-     */
-    public function testRollbackAttackDetection(): void
```
phenaproxima, line 1231: I moved this to `RollbackAttackTest`.

tests/Client/UpdaterTest.php
```diff
@@ -1408,29 +1132,10 @@ public function testMetadataFileTooBig(string $fixtureName, string $fileToChange
...
         // the known length, which should cause an exception during the update.
-        $this->serverStorage->fileContents[$fileToChange] = str_repeat('a', $knownLength + 1);
+        $this->serverFiles[$fileToChange] = str_repeat('a', $knownLength + 1);
 
         $this->expectException(DownloadSizeException::class);
         $this->expectExceptionMessage("Expected $fileToChange to be $knownLength bytes.");
-        $updater->refresh();
-    }
-
-    public function testSnapshotHashes(): void
```
phenaproxima, line 1418: This is now `SnapshotHashesTest`.

tests/Client/RollbackAttackTest.php
```diff
@@ -0,0 +1,65 @@
...
+             */
+            public function write(string $name, string $data): void
+            {
+                throw new \LogicException("Unexpected attempt to change client storage.");
+            }
+
+            /**
+             * {@inheritDoc}
+             */
+            public function delete(string $name): void
```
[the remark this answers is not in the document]
phenaproxima: This is doing the same thing that TestStorage is doing in HEAD. It overrides `delete()` to throw an exception. I merely decided not to repeat the exception itself, and just call `write()` internally.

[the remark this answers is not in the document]
tedbow: Since it still just 1 line it seems like would just be clearer to throw the exception in both places. I pretty sure I won't have asked the question if the exception was in both places

It doesn't seem worth adding comment instead of throwing the exception in 2 places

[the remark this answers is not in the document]
phenaproxima: Fixed!

tedbow (changes requested):

src/Client/Updater.php
```diff
@@ -73,12 +73,12 @@ class Updater
      *
      * @var \Tuf\Client\Repository
      */
-    private Repository $server;
+    protected Repository $server;
```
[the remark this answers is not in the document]
phenaproxima: Why?

Most of the other existing properties in `Updater` are protected. Things that are never overridden, like `$isRefreshed`, `$signatureVerifier`, even `$clock`. So what's the harm in just making `$repository` protected as well?

tests/Client/TestLoader.php
```diff
@@ -4,55 +4,47 @@
...
 use Tuf\Exception\RepoFileNotFound;
 use Tuf\Loader\LoaderInterface;
-use Tuf\Tests\TestHelpers\UtilsTrait;
 
-class TestLoader implements LoaderInterface
+/**
+ * Defines a trait to load files from a simulated, in-memory server.
+ *
+ * Classes using this trait should implement \Tuf\Loader\LoaderInterface.
+ */
```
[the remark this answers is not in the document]
phenaproxima: Happy to do it, but only if you want to keep this as its own separate class. If we're going back to a trait, then this should stay as-is. It's up to you.

tedbow (changes requested):
Still reviewing but had a few questions

tests/ClientTestBase.php
```diff
@@ -0,0 +1,153 @@
...
+     * @see ::loadClientAndServerFilesFromFixture()
+     */
+    protected TestLoader $serverFiles;
+
+    /**
+     * The server-side TUF repository.
+     *
+     * @var \Tuf\Tests\TestHelpers\TestRepository
+     */
+    protected TestRepository $server;
```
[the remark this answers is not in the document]
phenaproxima: `$server` is terribly named and I'm definitely open to changing that. However, it is NOT the files on the server. It is a higher level abstraction -- it's the untrusted metadata objects loaded from the server.

`$serverFiles`, on the other hand, is the actual raw files that are on the server.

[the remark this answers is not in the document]
tedbow: Maybe `$serverMetadata` ?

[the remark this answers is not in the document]
phenaproxima: Sure, that's definitely better.

tests/Client/TestLoader.php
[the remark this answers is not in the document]
phenaproxima: I get the sense you prefer the current way, so I've adjusted the comment.

tedbow (commented):
Ok the only test from the old version I could NOT find is `testUnsupportedRepo`. where is that one?

tedbow:
besides my last comment I think looks good. Thanks for the patience and detailed explanations

tedbow:
Ok I guess

> 

I guess this was moved to `\Tuf\Tests\Client\PathHashPrefixesTest::test`

tedbow (approved):
🎉 Great to get this in

PR #335, opened 2023-04-10
Support "classic" path hash prefixes for delegated roles

### LINKED ISSUE #191: Support `path_hash_prefixes` from the TUF Spec

phenaproxima:
This fixes #191 with explicit test coverage.


src/DelegatedRole.php
```diff
@@ -104,6 +104,16 @@ protected static function validate(array $data, Collection $constraints): void
      */
     public function matchesPath(string $target): bool
     {
+        if (isset($this->pathHashPrefixes)) {
+            $targetHash = hash('sha256', $target);
+
+            foreach ($this->pathHashPrefixes as $prefix) {
+                if (str_starts_with($targetHash, $prefix)) {
+                    return true;
```
tedbow, line 112: So if we didn't find the tart it would go on to `if ($this->paths) {` but that is ok because our metadata validation ensures they can never be both, correct?

phenaproxima: Yes.

tests/Client/HashedBinsTest.php
```diff
@@ -0,0 +1,23 @@
...
+        $this->loadClientAndServerFilesFromFixture('HashedBins');
+
+        // The fixture contains targets a.txt through z.txt (ASCII 97 through
+        // 122), distributed across hashed bins.
+        $targets = array_map(fn (int $c) => chr($c) . '.txt', range(97, 122));
+
+        $updater = $this->getUpdater();
+        // We should be able to download every single target without trouble.
+        foreach ($targets as $name) {
+            $this->assertSame("Contents: $name", $updater->download($name)->getContents());
```
tedbow, line 20: It looks like this is the only place that we rely on the fact that our python builder tool makes the contents with this pattern ""Contents: $name". Should we comment as to why we know this is the contents of the file?

tedbow (changes requested):
Looks good other than 1 suggested comment

tedbow (approved):
Just the one comment

PR #336, opened 2023-04-11
Remove all infrastructure that specifically forbids path_hash_prefixes

### LINKED ISSUE #191: Support `path_hash_prefixes` from the TUF Spec

phenaproxima:
This is the first PR of 3 to fix #191.

We have a lot of infrastructure related to specifically forbidding `path_hash_prefixes` in targets metadata. This PR removes all of that. It doesn't _add_ support for hash bins, but it removes the stuff that _disallows_ that support. Including:

* The UnsupportedDelegation fixture, whose only purpose was to confirm that hash bins are unsupported.
* PathHashPrefixesTest, for the same reason.
* Our specialized version of Symfony's Collection validation constraint, which implemented the concept of unsupported fields at the validation level, and related test coverage.


tedbow (approved):
Looks good!

PR #337, opened 2023-04-11
Allow DelegatedRole to validate and carry path_hash_prefixes

phenaproxima:
Part 2 of fixing #191.

This adds support for `path_hash_prefixes` to the DelegatedRole metadata object, and ensures it's mutually exclusive with `paths`.

### DIFF
--- a/src/DelegatedRole.php
+++ b/src/DelegatedRole.php
@@ -53,8 +53,8 @@ class DelegatedRole extends Role
             'paths' => new Optional([
                 new Type('array'),
                 new All([
-                   new Type('string'),
-                   new NotBlank(),
+                    new Type('string'),
+                    new NotBlank(),
                 ]),
             ]),
             'path_hash_prefixes' => new Optional([
--- a/src/DelegatedRole.php
+++ b/src/DelegatedRole.php
@@ -50,6 +50,8 @@ class DelegatedRole extends Role
                 ]
             ),
             'terminating' => new Required(new Type('boolean')),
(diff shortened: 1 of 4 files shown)


src/DelegatedRole.php
```diff
@@ -45,18 +50,46 @@ public static function createFromMetadata(iterable $roleInfo, string $name = nul
                 ]
             ),
             'terminating' => new Required(new Type('boolean')),
-            'paths' => new Required(new Type('array')),
+            'paths' => new Optional([
```
tedbow, line 53: We should  explain these are mutually exclusive and add a `@see` to `validate`

src/Metadata/TargetsMetadata.php
```diff
@@ -116,6 +116,13 @@ protected static function getSignedCollectionOptions(): array
                                     new NotBlank(),
                                 ]),
                             ]),
+                            'path_hash_prefixes' => new Optional([
```
tedbow, line 119: So they are not mutually exclusive here?

tedbow (changes requested):
1 comment suggestion and 1 question, and 1 question

src/Metadata/TargetsMetadata.php
phenaproxima: They are. But this information is never _used_ by TargetsMetadata, so it make more sense to do the validation in DelegatedRole, where it's actually meaningful.

tedbow (approved):
Looks good

PR #340, opened 2023-05-19
Indexed arrays with >= 10 items are incorrectly canonicalized

phenaproxima:
These arrays (which are detectable with PHP 8.1's `array_is_list()` function) should not have their keys sorted -- by definition, their keys are already sorted. But we sort them anyway, which results in them being incorrectly canonicalized, and thus causing false negatives during signature verification.

We should use Symfony's PHP 8.1 polyfill to bring in `array_is_list()`, and then ensure we don't sort the keys of indexed arrays when canonicalizing.

This bug manifests on arrays with more than 10 items, because keys like `10`, `11`, and so forth get put after 1, but before 2. So you end up with keys in the wrong order: `1, 10, 11, 2`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,8 @@
         "ext-json": "*",
         "paragonie/sodium_compat": "^1.13",
         "symfony/validator": "^4.4 || ^5 || ^6",
-        "guzzlehttp/psr7": "^2.4"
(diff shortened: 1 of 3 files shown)


tedbow (commented):
Looks good!

PR #342, opened 2023-05-19
Ensure indexed arrays are not sorted alphabetically at any level

phenaproxima:
Fixes #340 properly. That PR's fatal flaw is that it only worked for the top level data structure, rather than recursively.

### DIFF
--- a/tests/Unit/CanonicalJsonTraitTest.php
+++ b/tests/Unit/CanonicalJsonTraitTest.php
@@ -25,6 +25,17 @@ class CanonicalJsonTraitTest extends TestCase
     }
 
     /**
+     * Indexed arrays with >= 10 items should not be changed.
+     */
+    public function testIndexedArraysAreNotChanged(): void
+    {
+        $indexed = array_fill(0, 20, 'Hello!');
+        $this->assertTrue(array_is_list($indexed));
+        $canonicalized = static::encodeJson($indexed);
+        $this->assertSame($indexed, static::decodeJson($canonicalized));
+    }
+
+    /**
      * @covers ::encodeJson
      */
     public function testSlashEscaping(): void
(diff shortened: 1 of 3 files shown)


tests/Unit/CanonicalJsonTraitTest.php
```diff
@@ -24,6 +24,28 @@ public function testSort(): void
...
 
+    /**
+     * Indexed arrays with >= 10 items should not be changed.
+     */
+    public function testIndexedArraysAreNotChanged(): void
+    {
+        $original = [
+          'b' => array_fill(0, 20, 'Hello!'),
+          'a' => 'Canonically speaking, I go before b.',
+        ];
```
tedbow: Isn't the other thing about array_is_list is the numeric keys have to be consecutive?

should we test with this also?

$array = [
]

tests/Unit/CanonicalJsonTraitTest.php
```diff
@@ -22,6 +22,25 @@ public function testSort(): void
...
         $this->assertSame($unsortedData, $sortedData);
+
+        // Indexed arrays should not be sorted in alphabetical order, at any
+        // level.
+        $data = [
+            // This must have at least 10 items, since that will be sorted
+            // alphabetically (which is what we're trying to avoid).
+            'b' => array_fill(0, 20, 'Hello!'),
+            'a' => 'Canonically speaking, I go before b.',
+        ];
```
tedbow, line 33: Isn't the other thing about array_is_list is the numeric keys have to be consecutive?

should we test with this also?

```
$not_list = [
          3 => 'three',
          2 => 'two'
        ]
```

This would sorted, right?

tedbow (changes requested):

tests/Unit/CanonicalJsonTraitTest.php
phenaproxima: Yeah, the array does need to have consecutive integer keys, or it's not considered a list.

tests/Unit/CanonicalJsonTraitTest.php
phenaproxima: I think so, yes. I think it can't hurt to add that coverage.

tedbow (approved):
Looks good!

PR #343, opened 2023-05-19
TargetsMetadata::toCanonicalJson() can try to access an undefined key

phenaproxima:
If a TargetsMetadata object has no `delegations` key, `toCanonicalJson()` will still trigger an "Undefined key `delegations`" error from this bit:

```php
$normalized['delegations']['keys'] = (object) $normalized['delegations']['keys'];
```

We should ensure that $normalized has a `delegations` key before doing this.


tedbow (approved):
Looks good

PR #347, opened 2023-07-01
Re-add recurive calls for list arrays

SniperSister:
The fix for #340 prevented the correct sorting of array keys of child arrays if the parent array is list array. This PR fixes the issue and adds a specific test case.


phenaproxima:
This was fixed in #366. :) 🎉

PR #348, opened 2023-07-01
Fix invalid canonical JSON of TargetsMetadata

SniperSister:
We ran into an issue where php-tuf was unable to verify a signature of a valid targets metadata file.

After some closer investigation we realized that this was caused by a mismatch between the canonical JSON that is generated by php-tuf and the canonical JSON that the CLI tool generated for the signature.

The root cause for is the fact that "toNormalizedArray" in TargetsMetadata converted multiple properties from an array to object. Because of that, "sortKey" in the CanonicalJsonTrait did not correctly recurse into the sublevels, leaving array keys in sublevels unsorted. That leads to an invalid canonical JSON.

In this PR we removed the toNormalizedArray method and instead decided to override the toCanonicalJSON method. That also is a better fit for this specific usecase, as the casting to object isn't meant to normalize the array itself but specifically modifiy the generated JSON output.

### DIFF
--- a/tests/Metadata/TargetsMetadataTest.php
+++ b/tests/Metadata/TargetsMetadataTest.php
@@ -212,4 +212,77 @@ class TargetsMetadataTest extends MetadataBaseTest
         $this->expectExceptionMessage('Delegated role names must be unique.');
         static::callCreateFromJson($json);
     }
+
+    public function testKeysAreSorted(): void
+    {
+        $data = [
+            'signatures' => [],
+            'signed' => [
+                '_type' => 'targets',
+                'version' => 1,
+                'targets' => [
+                    'foo' => [
+                        'custom' => [
+                            'z' => 'Yes',
+                            'a' => 'No',
+                        ],
+                    ],
+                    'baz' => [],
+                ],
+                'delegations' => [
+                    'keys' => [
+                        'z' => [],
+                        'a' => [],
+                    ],
+                ],
+            ],
+        ];
+        $metadata = new TargetsMetadata($data, '');
+
+        $decoded = static::decodeJson($metadata->toCanonicalJson());
+        // The out-of-order keys should have been reordered.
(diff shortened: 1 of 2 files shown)


SniperSister:
Any chance of getting that merged :) @phenaproxima

phenaproxima:
Hi, I'm really sorry this has taken so long to review. I'd like to see this get fixed, but I think we'll need some kind of test coverage (I can take a crack at that), and I'm not sure all the changes in the PR are strictly necessary to fix the bug. I'll have to dig more deeply into it.

Can you provide me with an example of the "wrong" JSON being created by PHP-TUF, and what it _should_ be instead?

phenaproxima:
I think it makes sense to simplify this PR a bit, by just having `toNormalizedArray()` sort keys first.

I also agree that it makes sense to remove `toNormalizedArray()` entirely, but that's for a separate PR, not this one.

tests/Metadata/TargetsMetadataTest.php
```diff
@@ -212,4 +212,68 @@ public function testDuplicateDelegatedRoleNames(): void
         $this->expectExceptionMessage('Delegated role names must be unique.');
         static::callCreateFromJson($json);
     }
+
+    public function testKeysAreSorted(): void
+    {
+        $data = [
+            'signatures' => [],
+            'signed' => [
```
tedbow, line 220: Why not have top level keys out of order and assert they are fixed too?

tedbow (changes requested):
1 idea but looks good

tests/Metadata/TargetsMetadataTest.php
phenaproxima: Sure, why not? Added an assertion to that effect.

tedbow (approved):
Thanks, looks good

PR #354, opened 2023-10-27
Adopt static caching

phenaproxima:
Our Composer integration plugin currently has _terrible_ performance. All php-tuf/composer-integration#87 does is reduce our network traffic (which, to be clear, is a good thing). We're still talking to the server _a lot_, and therefore blocking asynchronous operations.

I think we should add static caching. Because here's the thing -- do we really expect data downloaded from the server to change _in the middle of a Composer operation_? And, if it does, isn't that bad? Wouldn't it lead to inconsistency and failed validation, if the TUF metadata we've loaded _into memory_ no longer matches what's coming from the server?

There are two places where I think static caching will give us a much-needed performance boost:

1. Loading data from the server. If we've already loaded something, then we should maintain it in memory, both for performance and consistency. If it needs to be refreshed from the server, well...that'll happen the next time the Composer plugin runs, since the static cache only lives in memory and has no persistence beyond the current PHP process. We can accomplish this by implementing a simple `LoaderInterface` decorator that implements a static cache, which the plugin should then take advantage of explicitly.
2. Figuring out which targets metadata contains the information about a particular target. If we have to search delegated roles, this can take an awful long time, and trigger another huge round of pinging the server. This should be an internal static cache of `getMetadataForTarget()` that maps a particular target name to an already-loaded metadata object.

Now, to be clear, both of these things could be done within the plugin itself. We don't _have_ to do this in PHP-TUF. But IMHO it's good practice and we should provide it.

### DIFF
--- a/src/Client/Updater.php
+++ b/src/Client/Updater.php
@@ -348,20 +348,26 @@ class Updater
      *
      * @param string $target
      *   The path of the target file.
+     * @param bool $useCache
+     *   (optional) Whether to load the metadata from an internal static cache.
      *
      * @return \Tuf\Metadata\TargetsMetadata|null
      *   The targets metadata with information about the desired target, or null if no relevant metadata is found.
      */
-    protected function getMetadataForTarget(string $target): ?TargetsMetadata
+    protected function getMetadataForTarget(string $target, bool $useCache = true): ?TargetsMetadata
     {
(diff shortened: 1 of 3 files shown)


phenaproxima:
Superseded by https://github.com/php-tuf/composer-integration/pull/89.

PR #358, opened 2023-12-06
Filter irrelevant delegated roles up front

phenaproxima:
As part of the project to make delegation searches parallel, we need to identify the roles we want to search up-front, rather than iteratively. So, let's do that -- we should only search roles that match the path of the target, up to the first terminating role. This means we can also remove a level of nesting from our code.

### DIFF
--- a/src/Client/Updater.php
+++ b/src/Client/Updater.php
@@ -428,7 +428,21 @@ class Updater
         foreach ($targetsMetadata->getDelegatedKeys() as $keyId => $delegatedKey) {
             $this->signatureVerifier->addKey($keyId, $delegatedKey);
         }
+
+        $delegatedRoles = [];
         foreach ($targetsMetadata->getDelegatedRoles() as $delegatedRole) {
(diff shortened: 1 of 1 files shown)


src/Client/Updater.php
```diff
@@ -428,7 +428,21 @@ private function searchDelegatedRolesForTarget(TargetsMetadata $targetsMetadata,
...
+
+                if ($delegatedRole->isTerminating()) {
+                    break;
+                }
+            }
+        }
+
+        foreach ($delegatedRoles as $delegatedRole) {
             $delegatedRoleName = $delegatedRole->getName();
             if (in_array($delegatedRoleName, $searchedRoles, true)) {
```
tedbow, line 447: should we move this to loop above. all tests still pass

phenaproxima: We could, but I think we might have to change it later when I convert this to a set of promises, rather than an iterative loop.

PR #359, opened 2023-12-06
Prevent path traversal by URL-encoding the names of all saved metadata

phenaproxima:
Right now, we save trusted metadata using whatever role name it has. If the role name has path traversal components (for example, `parent_role/../grandparent_role`), the file would be saved as `path/to/metadata/parent_role/../grandparent_role.json`. That's unauthorized filesystem access.

To fix this, the role name should be URL-encoded. To continue the previous example, it should be saved as `path/to/metadata/parent_role%2F..%2Fgrandparent_role.json` to prevent path traversal.

### DIFF
--- a/tests/Unit/FileStorageTest.php
+++ b/tests/Unit/FileStorageTest.php
@@ -147,7 +147,7 @@ class FileStorageTest extends TestCase
         $metadata->ensureIsTrusted()->shouldBeCalled();
         $storage->save($metadata->reveal());
 
-        $expectedFileName = urlencode($roleName) . '.json';
+        $expectedFileName = 'hello%2F..%2Fthere%21.json';
         $this->assertFileExists($dir . '/' . $expectedFileName);
 
         $method = new \ReflectionMethod($storage, 'read');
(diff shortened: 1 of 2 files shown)


tests/Unit/FileStorageTest.php
```diff
@@ -131,4 +132,26 @@ public function testWriteMetadata(string $metadataClass, string $role, string $e
...
+        $dir = sys_get_temp_dir();
+        $storage = new FileStorage($dir);
+
+        $metadata = $this->prophesize(MetadataBase::class);
+        $metadata->getRole()->willReturn($roleName);
+        $metadata->getSource()->willReturn($fileContents);
+        $metadata->ensureIsTrusted()->shouldBeCalled();
+        $storage->save($metadata->reveal());
+
+        $expectedFileName = urlencode($roleName) . '.json';
```
wimleers, line 150: IMHO this is much clearer:
```suggestion
        $expectedFileName = 'hello%2F..%2Fthere.json';
```

(Whenever possible, make test expectations have no logic, so that the before vs after is clearly visible.)

wimleers (changes requested):

tedbow (approved):
Looks good, thanks!

PR #361, opened 2024-02-07
Drop compatibility with Guzzle 6

phenaproxima:
We're not getting anything out of Guzzle 6 except for extra complexity, as well as having to walk carefully around the minefield of compatibility between two different _major_ versions of `guzzlehttp/promises`, which is a library we actually _do_ depend on. Let's drop compatibility with Guzzle 6.

For Drupal's part, it hasn't been compatible with Guzzle 6 since Drupal 9, which is now EOL.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
         "squizlabs/php_codesniffer": "^3.7",
         "slevomat/coding-standard": "^8.2",
         "phpspec/prophecy-phpunit": "^2",
-        "guzzlehttp/guzzle": "^6.5 || ^7.2",
+        "guzzlehttp/guzzle": "^7.2",
         "phpspec/prophecy": "^1.16"
     },
     "license": "MIT",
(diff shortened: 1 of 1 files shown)


tedbow:
re:

> We're not getting anything out of Guzzle 6 except for extra complexity, 
Is there any extra complexity in code we now have because are supporting Guzzle 6? If so should remove it here?

phenaproxima:
None that I know of, because Guzzle itself is only a dev dependency. However, trying to manage the dependency itself adds extra pain in some situations.

tedbow (approved):
Looks good

phenaproxima:
Okay...if I search the codebase for `GuzzleHttp`, I'm not seeing us using anything except `guzzlehttp/psr7` and `guzzlehttp/promises`. Maybe we can completely drop the dependency on `guzzlehttp/guzzle`.

phenaproxima:
For archaeological purposes, the Guzzle dependency was added in #272.

PR #363, opened 2024-02-07
Drop support for PHP 8.0

tedbow:
PHP 8.0 is no longer supported


tedbow:
Talked with @phenaproxima  about this. We can take up any advantages from dropping 8.0 later in follow-ups

PR #366, opened 2024-03-08
Replace outdated Python TUF tool with our own fixture generation code

phenaproxima:
Since the inception of this project, we have been reliant on the Python reference implementation of TUF.

The problem is that we've been stuck on a pre-release version of it, because we _cannot_ update to the actual releases. The reason we can't update is because we are relying on the Python tool's repository builder in order to create our test fixtures...and that repository builder has been completely removed and won't be coming back. So the Python dependency, for our purposes, is a permanently dead end.

We need some way to generate complete TUF repositories in order for our tests to work. I think we should do it ourselves, in PHP.

This PR adds a PHP-based fixture builder that generates the _exact same_ TUF metadata as the Python tool did, and uses it to build our fixtures. It also adapts our tests to use the PHP-built fixtures, and removes our Python dependency.


tests/FixtureBuilder/Key.php
```diff
@@ -0,0 +1,150 @@
...
+            'keyid_hash_algorithms' => ['sha256', 'sha512'],
+        ];
+    }
+
+    public function id(): string
+    {
+        return hash('sha256', self::encodeJson($this->toArray()));
+    }
+
+    public static function fromStaticList(): self
```
tedbow, line 55: could you add a doc comment and possibly some inline comments

tedbow (dismissed):
changes under, /tests and /src generally look good. 

I noticed `tests/FixtureBuilder` has almost no function doc blocks and some `{@inheritDoc}` that where the overridden function has no doc block.

So I think we need some more comments. Not saying every single function in needs a comment but we some

phenaproxima:
> How will this affect https://github.com/php-tuf/composer-integration ?

It shouldn't. The fixture paths are all the same, and except for the keys and expiration dates (which the plugin doesn't, for testing purposes, care about), the files themselves are identical.

tests/FixtureBuilder/Key.php
phenaproxima: Yup! Added comments where they seemed necessary.

tests/FixtureBuilder/Fixture.php
```diff
@@ -0,0 +1,275 @@
...
+        for ($x = 0; $x < $prefixCount; $x += $binSize) {
+            $high = $x + $binSize - 1;
+
+            if ($x === $high) {
+                $name = sprintf("%0{$prefixLength}x", $x);
+            } else {
+                $name = sprintf("%0{$prefixLength}x-%0{$prefixLength}x", $x, $high);
+            }
+
+            if ($binSize === 1) {
```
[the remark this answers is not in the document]
phenaproxima: Oooooh...great catch. I changed this to `=== 1.0`, and PHPStorm has stopped complaining.

tests/FixtureBuilder/Fixture.php
```diff
@@ -0,0 +1,275 @@
+<?php
+
+namespace Tuf\Tests\FixtureBuilder;
+
+use Symfony\Component\Filesystem\Filesystem;
+
+/**
+ * This is our in-house TUF fixture builder.
```
tedbow, line 8: I don't think we need to specify "in-house" because anybody new to the codebase would have to have the context that we use to use the Python TUF for this to make sense.

tests/FixtureBuilder/Payload.php
```diff
@@ -0,0 +1,169 @@
...
+ * A payload represents a single role in the TUF repository. This could be one
+ * of the four top-level roles (root, timestamp, snapshot, and targets), or it
+ * could be a delegated targets role at any level of nesting.
+ *
+ * Every payload has relationships to other payloads. There are two kinds
+ * of relationships:
+ *
+ * - A "parent" relationship is for a payload that enumerates other payloads.
+ *   For example, a snapshot has a parent relationship to all targets roles,
+ *   regardless of nesting level. The parent may needs to be updated when one
```
tedbow, line 21:
```suggestion
 *   regardless of nesting level. The parent may need to be updated when one
```
Not sure if should be `may need` or just `needs`

tests/FixtureBuilder/Payload.php
```diff
@@ -0,0 +1,169 @@
...
+ * A payload is a builder for a server-side TUF metadata file.
+ *
+ * A payload represents a single role in the TUF repository. This could be one
+ * of the four top-level roles (root, timestamp, snapshot, and targets), or it
+ * could be a delegated targets role at any level of nesting.
+ *
+ * Every payload has relationships to other payloads. There are two kinds
+ * of relationships:
+ *
+ * - A "parent" relationship is for a payload that enumerates other payloads.
```
tedbow, line 19: Should the docs for parent and key ring just `@param` docs for `__construct`?

phenaproxima: I thought about that, but I don't think that's really the best place for it. This class-level doc block is more general, and needs to explain the overall choices of how these classes are meant to fit together.

tests/FixtureBuilder/Fixture.php
```diff
@@ -0,0 +1,275 @@
...
+        ]);
+        $this->timestamp = new Timestamp($this->root, $expires, [
+            Key::fromStaticList(),
+        ]);
+
+        $this->snapshot = new Snapshot($this->root, $this->timestamp, $expires, [
+            Key::fromStaticList(),
+        ]);
+        // Disable hashes and file sizes in the snapshot metadata, which matches
+        // the default configuration of the Python TUF tool.
```
tedbow, line 67: I think we should state reason we want this to be this way instead of referring to the python tool.

tedbow:
I see some references to the "python tool" I think if we need those then somewhere in a class doc, maybe `\Tuf\Tests\FixtureBuilder\Fixture` we should reference the exact version of the tool we were using before. Otherwise someone new to the codebase would have to no way of checking back to see why did things that way or if we got it wrong.

tedbow:
> To prevent it from breaking, one thing we could do is copy the Python stuff into the plugin for now, so that it still has the Python infrastructure (just not coming from this package). That is something we could do before merging this, to keep things running smoothly.

Well right now require ` "php-tuf/php-tuf": "0.1.3",` specifically so we should be ok I think

phenaproxima:
Yes. I put it at the top of `Fixture`.

README.md
tedbow, line 1: Should remove some of the links under `Resources` to the Python tool? Maybe we just need 1 link now?

phenaproxima:
> > // This field is not part of the TUF specification and is being
> > // removed from the Python TUF reference implementation in
> > // [theupdateframework/python-tuf#848](https://github.com/theupdateframework/python-tuf/issues/848).
> > // If it is provided though we only support the default value which
> > // is passed on from a setting in the Python `securesystemslib`
> > // library.
> 
> from: \Tuf\Metadata\ConstraintsTrait::getKeyConstraints Should make changes based on that fact that we aren't using this anymore?

Sure, but that's not in scope for this PR. It would change the generated fixtures and possibly require test changes.

README.md
phenaproxima: I thought about that, but since Python continues to be the reference implementation of TUF -- despite not having a repository generator anymore -- I think we can keep those links.

tedbow (approved):
Thanks @phenaproxima

PR #367, opened 2024-03-08
Replace simple metadata getters with read-only properties

phenaproxima:
Addresses #60.

`public readonly` properties are a great choice to replace these simple getters. Even for arrays (i.e., `getSigned() -> $signed`) -- you can't modify any level of a read-only array. 🔒 🎉 

This lets us remove boilerplate, increase readability, and improve maintainability.


tedbow (approved):
Tests are still running but looks good if they pass

PR #368, opened 2024-03-09
Use Symfony's Unique constraint where appropriate

### LINKED ISSUE #317: Use Symfony 6.1's Unique constraint
In #314, we added some validation that would be a perfect fit for Symfony's `Unique` validation constraint. However, we couldn't use it because we'd need to use the constraint's `fields` option, which was only added in Symfony 6.1 (see https://symfony.com/doc/current/reference/constraints/Unique.html#fields). Adopting Symfony 6.1 would necessitate requiring PHP 8.1 at a minimum, and dropping support for Symfony 4 and 5. So, for now, we're just doing custom validation.

But, when we're ready to require PHP 8.1, maybe we can require Symfony 6.1 as well, and then refactor the validation to use the `Unique` constraint.

phenaproxima:
This fixes #317.

It also removes the redundant use of `new Type('array')` in places where we're also using `new All([...])`. [The `All` validator ensures that it's given an array or a `Traversable`](https://github.com/symfony/validator/blob/6.1/Constraints/AllValidator.php#L37), which means we don't also need to specify `new Type('array')`. (Maybe we did in earlier versions of Symfony Validator, but it's not necessary in 6.1.)

### DIFF
--- a/tests/Metadata/MetaDataBaseTest.php
+++ b/tests/Metadata/MetaDataBaseTest.php
@@ -312,6 +312,7 @@ abstract class MetadataBaseTest extends TestCase
                 $newValue = 'Abb';
                 break;
             case 'array':
+            case 'iterable':
                 $newValue = 3060;
                 break;
             case 'boolean':
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         "php": "^8.1",
         "ext-json": "*",
         "paragonie/sodium_compat": "^1.13",
-        "symfony/validator": "^6",
+        "symfony/validator": "^6.1",
         "guzzlehttp/psr7": "^2.4",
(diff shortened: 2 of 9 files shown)


tests/Metadata/MetaDataBaseTest.php
```diff
@@ -312,6 +312,7 @@ public function testInvalidField(string $expectedField, string $expectedType): v
                 $newValue = 'Abb';
                 break;
             case 'array':
+            case 'iterable':
```
phenaproxima, line 315: This is needed because `AllValidator` looks for iterables, which includes arrays, but when it errors, it says "must be an iterable".

tedbow (approved):
👍

PR #369, opened 2024-03-11
Completely remove support for keyid_hash_algorithms

phenaproxima:
The `keyid_hash_algorithms` field is no longer part of the TUF specification: https://theupdateframework.github.io/specification/latest/#file-formats-keys

We should remove support for it and the related test coverage.


tedbow (approved):
👍

PR #370, opened 2024-03-11
Test on PHP 8.3

phenaproxima:
PHP 8.3 has been released for a while now, so it should be in our testing matrix.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
       max-parallel: 10
       matrix:
         operating-system: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -14,14 +14,14 @@ env:
 
 jobs:
   build:
-    name: PHP ${{ matrix.php }} test on ${{ matrix.operating-system }} with ${{ matrix.dependencies }}
```
phenaproxima, line 17: Removed because this has not been a thing since we ceased supporting multiple versions of Symfony. So it was never actually being replaced with anything.

tedbow (approved):
👍

PR #372, opened 2024-03-11
Remove additional redundant uses of Symfony's Type constraint

phenaproxima:
In #368, we removed redundant uses of the `Type` constraint. I missed a few spots, though.


tedbow (approved):
👍

PR #373, opened 2024-03-12
Remove Symfony's PHP 8.1 polyfill from dependencies

phenaproxima:
We depend directly on PHP 8.1 or later, so we don't need this polyfill.


tedbow (approved):
Looks good!

PR #374, opened 2024-03-12
Completely remove Prophecy

phenaproxima:
We are only using Prophecy in two places in our tests; the rest of the time, we're using PHPUnit's mocking framework. Since Prophecy necessitates additional dependencies and has some incompatibilities with PHP 8.2 (you can't set a mock's constructor arguments without triggering a language-level deprecation), I don't think we should be using two mocking frameworks. Let's remove Prophecy and rely only on PHPUnit.


tedbow (approved):
👍

PR #376, opened 2024-03-12
Remove every instance of `@return void`

phenaproxima:
`@return void` is a contradiction in terms and it has always irked me that our code base is littered with them. This PR removes all of them.


tedbow (approved):
👍

PR #377, opened 2024-03-12
Remove DelegatedRole::isTerminating() in favor of read-only property

phenaproxima:
Forgot to do this as part of #367. :)


tedbow (approved):
Looks good

PR #378, opened 2024-03-13
Consistently use CanonicalJsonTrait in our code base

phenaproxima:
Our code base is littered with examples of us calling `json_encode()` and `json_decode()` directly, even in places that use `CanonicalJsonTrait`.

Considering how it's very important than JSON be encoded and decoded consistently (one might even say canonically) in PHP-TUF, it makes sense to always use the `CanonicalJsonTrait` methods unless there's a specific reason not to.


tedbow (approved):
👍

PR #379, opened 2024-03-13
Replace RepoFileNotFound with NotFoundException

phenaproxima:
We have two exceptions, `RepoFileNotFound` and `NotFoundException`, which serve very similar purposes. We should consolidate them. This converts all usages of `RepoFileNotFound` to `NotFoundException`, and removes `RepoFileNotFound`.


.github/workflows/build.yml
```diff
@@ -68,7 +68,7 @@ jobs:
         run: composer test
 
       - name: Check dependencies for known security vulnerabilities
-        run: composer audit
+        run: composer audit --no-dev
```
tedbow, line 71: How is this related?

tedbow (approved):
Looks good

.github/workflows/build.yml
phenaproxima: Proof of failure: https://github.com/php-tuf/php-tuf/actions/runs/8266701537/job/22615313427#step:13:1

PR #380, opened 2024-07-01
Allow all CI workflows to be triggered manually

phenaproxima:
I need this now, so that I can confirm that PHP-TUF is working before I start hacking on the plugin again.


tedbow (approved):
🎉

PR #381, opened 2024-07-02
Update TUF spec to 1.0.33

### LINKED ISSUE #352: TUF specification has a new version - v1.0.33
Hey, it seems there's a newer version of the TUF specification - [v1.0.33](https://github.com/theupdateframework/specification/blob/v1.0.33/tuf-spec.md)

The version which [php-tuf/php-tuf](https://github.com/php-tuf/php-tuf) states it supports is - [v1.0.32](https://github.com/theupdateframework/specification/blob/v1.0.32/tuf-spec.md)

The following is a comparison of what changed between the two versions - [Compare v1.0.32 to v1.0.33](https://github.com/theupdateframework/specification/compare/v1.0.32...v1.0.33)

Please review the newer version and address the changes.

phenaproxima:
Fixes #352.

I had to carefully read the spec difference (https://github.com/theupdateframework/specification/compare/v1.0.32...v1.0.33) to determine that, in fact, we're already compliant, based on this code in `SignatureVerifier::checkSignatures()`:

```
            if ($role->isKeyIdAcceptable($signature['keyid']) && $this->verifySingleSignature($metadata->toCanonicalJson(), $signature)) {
                $verifiedKeySignatures[$signature['keyid']] = true;
            }
```

If I understand it correctly, the spec is simply being explicit that the same key cannot count more than once. We were already doing that.

So I think we're good here!

### DIFF
--- a/.github/workflows/specification-version-check.yml
+++ b/.github/workflows/specification-version-check.yml
@@ -11,4 +11,4 @@ jobs:
       issues: write
     uses: theupdateframework/specification/.github/workflows/check-latest-spec-version.yml@master
     with:
-      tuf-version: "v1.0.32" # Should be updated to the version the project supports either manually or extracted automatically. You can see how python-tuf did that as an example.
+      tuf-version: "v1.0.33" # Should be updated to the version the project supports either manually or extracted automatically. You can see how python-tuf did that as an example.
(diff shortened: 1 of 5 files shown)


tedbow (approved):
🎉

PR #382, opened 2024-07-03
Temporarily restore keyid_hash_algorithms for backwards compatibility with securesystemslib

phenaproxima:
[securesystemslib still includes `keyid_hash_algorithms` when canonicalizing keys](https://github.com/secure-systems-lab/securesystemslib/blob/release-0.30.0/securesystemslib/keys.py#L435), which puts it at odds with the TUF spec, which removed all mention of `keyid_hash_algorithms`  (and which we removed accordingly in #369).

Rugged, and possibly other server-side TUF repository generators, [are using securesystemslib and therefore including `keyid_hash_algorithms` when key IDs are computed](https://gitlab.com/rugged/rugged/-/blob/main/rugged/commands/lib/verification_keys.py#L67). We need to be compatible with that. So, although it is not spec-compliant, we can hotfix this on our end until the correct fix lands in Rugged, which is for it to _explicitly_ remove `keyid_hash_algorithms` from the canonicalized key before hashing it.

### DIFF
--- a/src/Key.php
+++ b/src/Key.php
@@ -74,8 +74,8 @@ final class Key
             ],
             // This field is no longer part of the TUF spec, but it's here for backwards compatibility
             // with securesystemslib 0.30.0 and older (a Python package used by Rugged, and possibly
-            // others, to generate server-side TUF metadata). We can remove this once ### is fixed
-            // and deployed.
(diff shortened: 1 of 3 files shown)


tedbow (approved):
Looks if we get the rugged issue linked.

PR #386, opened 2025-05-01
Avoid recursively looking up the same delegated roles multiple times

catch56:
Updater::searchDelegatedRolesForTarget() calls itself recursively, however it calculates the full list of delegated roles for 'bins' every time. This is causing severe CPU and memory issues when testing locally with Drupal core.

An improvement is to store the delegated roles in a class property.

For me this leads to approximately a 50% reduction in both function calls and peak memory usage when profiling composer update with xhprof. It also removes about 15 http requests from Guzzle, because it was doing the full http request and everything else to build the list for 'bins' each time.

Before:

Screenshot from 2025-05-01 12-05-48

Screenshot from 2025-05-01 12-07-57

After:

Screenshot from 2025-05-01 12-13-21

Screenshot from 2025-05-01 12-14-06

### DIFF
--- a/src/Client/Updater.php
+++ b/src/Client/Updater.php
@@ -385,7 +385,7 @@ class Updater
     private function fetchAndVerifyTargetsMetadata(string $role): TargetsMetadata
     {
         if (isset($this->targetsMetadata[$role])) {
-          return $this->targetsMetadata[$role];
+            return $this->targetsMetadata[$role];
         }
         $fileInfo = $this->storage->getSnapshot()->getFileMetaInfo("$role.json");
         // § 5.6.1
@@ -438,7 +438,7 @@ class Updater
             $this->signatureVerifier->addKey($keyId, $delegatedKey);
         }
         if (!isset($this->delegatedRoles[$targetsMetadata])) {
-          $this->delegatedRoles[$targetsMetadata] = $targetsMetadata->getDelegatedRoles();
+            $this->delegatedRoles[$targetsMetadata] = $targetsMetadata->getDelegatedRoles();
         }
 
         $delegatedRoles = [];
@@ -455,7 +455,7 @@ class Updater
         }
 
         foreach ($delegatedRoles as $delegatedRole) {
-          $delegatedRoleName = $delegatedRole->name;
+            $delegatedRoleName = $delegatedRole->name;
             if (in_array($delegatedRoleName, $searchedRoles, true)) {
                 // § 5.6.7.1
                 // If this role has been visited before, skip it (to avoid cycles in the delegation graph).
--- a/src/Client/Updater.php
+++ b/src/Client/Updater.php
@@ -76,6 +76,9 @@ class Updater
      */
     protected Repository $server;
 
+    /**
+     * @var array<string, PromiseInterface<TargetsMetadata>>
+     */
     private array $targetsMetadata = [];
 
     private \SplObjectStorage $delegatedRoles;
@@ -464,7 +467,8 @@ class Updater
 
             $this->signatureVerifier->addRole($delegatedRole);
             /** @var \Tuf\Metadata\TargetsMetadata $delegatedTargetsMetadata */
-            $delegatedTargetsMetadata = $this->fetchAndVerifyTargetsMetadata($delegatedRoleName)->wait();
+            $delegatedTargetsMetadata = $this->fetchAndVerifyTargetsMetadata($delegatedRoleName)
+                ->wait();
             if ($delegatedTargetsMetadata->hasTarget($target)) {
                 return $delegatedTargetsMetadata;
             }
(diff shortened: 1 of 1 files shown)


src/Client/Updater.php
```diff
@@ -417,19 +417,24 @@ private function getUpdateStartTime(): \DateTimeImmutable
...
      */
-    private function searchDelegatedRolesForTarget(TargetsMetadata $targetsMetadata, string $target, array $searchedRoles, bool &$terminated = false): ?TargetsMetadata
+    private function searchDelegatedRolesForTarget(TargetsMetadata $targetsMetadata, string $target, array $searchedRoles, bool &$terminated = false, array $allDelegatedRoles = []): ?TargetsMetadata
     {
         foreach ($targetsMetadata->getDelegatedKeys() as $keyId => $delegatedKey) {
             $this->signatureVerifier->addKey($keyId, $delegatedKey);
         }
+        if (!\count($allDelegatedRoles)) {
+            $allDelegatedRoles = $targetsMetadata->getDelegatedRoles();
+        }
```
tedbow, line 434: It has been a while since I looked at this but is this safe to do?. If we call `getDelegatedRoles()` once for the first `$targetsMetadata` but then in recursive calls when have a different `$delegatedTargetsMetadata` use the same `$allDelegatedRoles` won't this not work for any nested delegations defined in the delegated targets metadata itself.

catch56:
New before and after.


Before:
Screenshot from 2025-05-02 18-58-58


After:
Screenshot from 2025-05-02 18-56-17

src/Client/Updater.php
catch56: Pushed a new commit that adds a class property to save recalculation instead of passing things recursively.

catch56:
Better before after - some work moves around so the function detail isn't that useful, this is the top level xhprof summary:

Before:
Screenshot from 2025-05-02 20-33-40


After:

Screenshot from 2025-05-02 20-33-25

src/Client/Updater.php
```diff
@@ -76,6 +76,10 @@ class Updater
      */
     protected Repository $server;
 
+    protected array $targetsMetadata = [];
+
+    protected \SplObjectStorage $delegatedRoles;
```
phenaproxima, line 81:
```suggestion
    private \SplObjectStorage $delegatedRoles;
```

src/Client/Updater.php
```diff
@@ -76,6 +76,10 @@ class Updater
      */
     protected Repository $server;
 
+    protected array $targetsMetadata = [];
```
phenaproxima, line 79:
```suggestion
    private array $targetsMetadata = [];
```

src/Client/Updater.php
```diff
@@ -377,21 +382,26 @@ protected function getMetadataForTarget(string $target): ?TargetsMetadata
      * @return \GuzzleHttp\Promise\PromiseInterface<\Tuf\Metadata\TargetsMetadata>
      *   A promise wrapping the verified metadata for the role.
      */
-    private function fetchAndVerifyTargetsMetadata(string $role): PromiseInterface
+    private function fetchAndVerifyTargetsMetadata(string $role): TargetsMetadata
     {
+        if (isset($this->targetsMetadata[$role])) {
+            return $this->targetsMetadata[$role];
+        }
```
phenaproxima, line 389: I don't think we should change the return type here; let's just return an already-fulfilled promise. Or better yet, we can cache the promises; that way, even if a promise hasn't completed, we know we have a promise pending for that role.

src/Client/Updater.php
```diff
@@ -427,9 +437,12 @@ private function searchDelegatedRolesForTarget(TargetsMetadata $targetsMetadata,
         foreach ($targetsMetadata->getDelegatedKeys() as $keyId => $delegatedKey) {
             $this->signatureVerifier->addKey($keyId, $delegatedKey);
         }
+        if (!isset($this->delegatedRoles[$targetsMetadata])) {
+            $this->delegatedRoles[$targetsMetadata] = $targetsMetadata->getDelegatedRoles();
```
phenaproxima, line 441: We can use null-coalescing assignment for this to remove the `if` check.

```suggestion
            $this->delegatedRoles[$targetsMetadata] ??= $targetsMetadata->getDelegatedRoles();
```

phenaproxima (dismissed):
The changes make sense to me; this is just adding more static caching. But I think we can reduce the number of changes by continuing to rely on promises (which also means we can lean more into async in the future).

src/Client/Updater.php
catch56: Switched to caching promises - this seems to be OK!

catch56:
Updated xhprof screenshot after #390 was merged, to show the two working together:

image

PR #391, opened 2025-05-29
Use a generator for TargetsMetadata::getGeneratedRoles()

catch56:
image

Before (any of the PRs trying to deal with this):

Screenshot from 2025-05-29 16-38-11

After (combined with #390):

image

The short version is, instead of calling DelegatedRole::createFromMetadata 8,227 times, we can call it about 30 times.


catch56:
This looked promising from a cpu/performance perspective until I noticed a warning locally, and then after I fixed that, it's less effective than https://github.com/php-tuf/php-tuf/pull/386. Maybe there's something here, but there's not yet.

PR #392, opened 2025-05-29
Test on PHP 8.4

phenaproxima:
PHP 8.4 has been out for aeons and we don't test on it. Let's fix that, and fix any deprecations arising from that.


phenaproxima:
This is failing tests, but the deprecations are due to changes in Symfony and therefore out of this PR's scope.

